PR #3, opened 2017-08-22
retries: extract B2RetryPolicy so it can be replaced.

certainmagic:
B2DefaultRetryPolicy implements the default policy which should be fine
for almost all sdk users.  So far, B2StorageClientImpl always uses the
default policy.  I'll work on that in an upcoming change.

Prior to this change, the B2Sleeper was allowed to throw after the
decision was made to sleep some and retry.  As part of extracting the
retry policy, I wanted to be sure that the retry policy would always
find out if the operation succeeded or failed and that wouldn't be the
case if the sleep threw.  I no longer allow the B2Sleeper to throw.
Instead, if it's interrupted, it just resets the thread's interrupted
flag and returns early.  We count on the interruption being noticed
elsewhere (such as in the network call to the B2 servers).

tested with "gradle build javadoc".


core/src/main/java/com/backblaze/b2/client/B2DefaultRetryPolicy.java
```diff
@@ -0,0 +1,62 @@
...
+package com.backblaze.b2.client;
+
+import com.backblaze.b2.client.exceptions.B2Exception;
+
+/**
+ * B2DefaultRetryPolicy implements the retry policy described in the B2
+ * documentation.  It should be reasonable and sufficient for almost every
+ * use of B2.
+ *
+ * Each attempted operation should have a unique instead of this class
```
bwbeach: s/instead/instance/

core/src/main/java/com/backblaze/b2/client/B2LargeFileUploader.java
```diff
@@ -225,23 +226,25 @@ private B2FileVersion uploadPartsAndFinish(B2FileVersion largeFileVersion,
...
+                    final B2UploadPartRequest partRequest = B2UploadPartRequest
+                            .builder(partSpec.partNumber,
+                                    new B2PartOfContentSource(request.getContentSource(), partSpec.start, partSpec.length))
+                            .build();
+
+                    final B2Part part = webifier.uploadPart(uploadPartUrlResponse, partRequest);
+                    uploadPartUrlCache.unget(uploadPartUrlResponse);
+                    return part;
+                },
+                new B2DefaultRetryPolicy());
```
bwbeach: Should this use the retry policy supplier?

core/src/main/java/com/backblaze/b2/client/B2DefaultRetryPolicy.java
certainmagic: fixed locally.

core/src/main/java/com/backblaze/b2/client/B2LargeFileUploader.java
certainmagic: yes.  good eye.  :)

right now, we always use the default retryer.  in an upcoming change, i'll make it configurable from the outside and i'll fix it here too.

PR #4, opened 2017-08-28
Upload progress

certainmagic:
This adds B2UploadListener and arranges to call it as progress is made on uploading both small files and large files.

### DIFF
--- a/core/src/main/java/com/backblaze/b2/client/B2LargeFileUploader.java
+++ b/core/src/main/java/com/backblaze/b2/client/B2LargeFileUploader.java
@@ -34,8 +34,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.function.Supplier;
 
-import static com.backblaze.b2.util.B2DateTimeUtil.ONE_SECOND_IN_MSECS;
-
 class B2LargeFileUploader {
     private final B2Retryer retryer;
     private final B2StorageClientWebifier webifier;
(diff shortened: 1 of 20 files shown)


core/src/main/java/com/backblaze/b2/client/structures/B2UploadProgress.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright 2017, Backblaze Inc. All Rights Reserved.
+ * License https://www.backblaze.com/using_b2_code.html
+ */
+package com.backblaze.b2.client.structures;
+
+/**
+ * The B2UploadProgress summarizes the status of an upload.
```
bwbeach: It's not clear to me how to interpret this structure.  Maybe I haven't read enough about the other classes?

Is this the progress for one part?  Does the state diagram in the enum class apply to each part?

core/src/main/java/com/backblaze/b2/client/B2LargeFileUploader.java
```diff
@@ -248,21 +253,173 @@ private B2FileVersion uploadPartsAndFinish(B2FileVersion largeFileVersion,
         return retryer.doRetry("b2_finish_large_file", accountAuthCache, () -> webifier.finishLargeFile(accountAuthCache.get(), finishRequest), new B2DefaultRetryPolicy());
     }
 
+    /**
+     * ProgressAdapter turns byte-level updates into
```
bwbeach: into ...?

PR #6, opened 2017-09-01
httpClient: add a thread that cleans up idle connections.

certainmagic:
this is based on the recommendation and sample code in
  https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html

ran B2Sample locally in the debugger to watch the thread get created, run, and shutdown.
did a "./gradlew clean build javadoc" for good measure (but the unit tests don't exercise the http client code)


certainmagic:
I hadn't heard much about wait() in years either, so when it was in the
sample code, i read about it and it seemed appropriate and more appropriate
than sleep().  :)

ttfn,
ab

PR #9, opened 2017-10-11
Percent-encode files names when downloading by name.

bwbeach:
Also, update the percent-encoding to match the sample code
from the B2 web site.  It's not necessary to encode "/", and
file names read better when it's not encoded.

Testing:
- unit tests


certainmagic:
thanks, @bwbeach

PR #11, opened 2017-10-17
largeFileUploader: use retrySupplier more consistently.

certainmagic:
Marj noticed that I was incorrectly using the DefaultRetryPolicy when
I should've been using the retrySupplier to use the implementation
provided by the SDK user.


certainmagic:
marj co-wrote this with me.  :)

PR #12, opened 2017-10-18
Add bucket type filtering

certainmagic:


certainmagic:
the servers support this now, so we can merge it to master.  :)

PR #13, opened 2017-10-27
largeFileUploader: catch RejectedExecutionException and wrap it with …

certainmagic:
…a B2Exception.

i prefer to throw exceptions as B2Exceptions instead of leting low-level
exceptions go through.  also, throwing my own exception gives me a chance
to suggest what might've happened.

tested with updated unit test.


certainmagic:
i also exercised it locally in B2.java by calling shutdown on the executor.  :)

PR #20, opened 2017-11-17
Add corsRules to bucket creation and update requests

spartan218:
Note: these still have to be null with the current API, until implemented on servers.


certainmagic:
the build failed because it's didn't have access to the magic secrets needed to upload the build results.  that's fine.  the rest of the build passed so, i'm going to merge this.

PR #22, opened 2017-11-22
Add corsRulesName to the CorsRule class and fixed test cases.

spartan218:
Also added a comment about the difference between
B2CreateBucketRequestReal and B2CreatBucketRequest.


certainmagic:
this only failed the build because pull requests don't have access to b2 to upload artifacts and my travis script tries anyway.  i should improve that!

PR #23, opened 2017-11-28
b2StorageClient: add getUploadUrl() and getUploadPartUrl()

certainmagic:
using these directly is mostly discouraged because the SDK
handles lots of authentication and retry logic for you in
uploadSmallFile() and uploadLargeFile() and it won't do
that if you just get these upload urls directly and use
them outside its logic.

that said, they could be useful if you needed to give an
upload URL and authToken to some code that can't use the SDK.


certainmagic:
thx for the review, brian.  :)

PR #25, opened 2018-03-02
httpClientFactory: explicitly set the user-agent on the HttpClient

certainmagic:
we compute default the HttpClient user-agent the same way HttpClient
does it and we keep the result as a constant, so that we can prevent
HttpClientBuilder doing that work every time we make an HttpClient.
(we make them and discard them constantly and HttpClient computing
the default is expensive because it opens a resource from its jar
and we'd like to avoid those unnecessary syscalls and unzipping.)

besides, we override the value later.  :)

### DIFF
--- a/httpclient/src/main/java/com/backblaze/b2/client/webApiHttpClient/HttpClientFactoryImpl.java
+++ b/httpclient/src/main/java/com/backblaze/b2/client/webApiHttpClient/HttpClientFactoryImpl.java
@@ -14,9 +14,11 @@ import org.apache.http.conn.socket.ConnectionSocketFactory;
 import org.apache.http.conn.socket.PlainConnectionSocketFactory;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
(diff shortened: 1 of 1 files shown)


httpclient/src/main/java/com/backblaze/b2/client/webApiHttpClient/HttpClientFactoryImpl.java
```diff
@@ -37,6 +39,20 @@
...
+     * This is the user-agent we should use on Apache HttpClient instances.
+     * If we do not set it, HttpClientBuilder will compute this every time
+     * it creates an HttpClient (and we do that A LOT).  That wouldn't
+     * be so bad, except that internally it gets its own version by opening
+     * a resource stream which involves opening a jar and using a ZipFile
+     * instance, etc, so it's a non-obvious amount of work.  (at least as
+     * of httpcomponents-client-4.5.2).  So, we do the work once and
+     * manually set the userAgent from the resulting constant.
+     */
+    private static final String APACHE_HTTP_CLIENT_USER_AGENT = VersionInfo.getUserAgent("Apache-HttpClient",
```
bwbeach: Do we want to include `b2-sdk-java` in the user agent?

Issue #42, opened 2018-06-06
Changes to support application keys


certainmagic:
The b2 api will be supporting restricted "[application keys](https://www.backblaze.com/b2/docs/application_keys.html)" soon.

To support that, we will be changing the B2AccountAuthorizer to no longer have the getAccountId() method because it might only have an applicationKeyId and an applicationKey.  Instead, we will get the accountId from the first successful account authorization response.  (We considered requiring users to always provide the accountId too when setting up the storage client, but we think that will result in worse user interfaces for humans because they will need to provide three things, instead of just two.)

B2StorageClient.getAccountId()'s signature will change to allow it to throw B2Exception, since the authorization might fail.  We don't have a way to measure how many b2-sdk-java users might be calling getAccountId() and whether it will be an issue for it to start throwing.  

I plan to bump the major version number of the SDK to reflect that these changes aren't necessarily source code compatible with previous versions of the SDK.

certainmagic:
done.

PR #43, opened 2018-06-06
These changes are in reference to application key work.

spartan218:
We used to memorize the accountId passed into the AccountAuthorizer,
however that doesn't work when using applicationKeyId's to authenticate.

So in order for the SDK to work with the new Application Keys
we now store the accountId in the B2AccountAuthorizationCache after
successfully authorizing.

The changes were removing the B2Authorizer getAccount() method,
and calling getAccountId() from the cache, either with retries or not.

Updated Tests, passes All Tests in core and works with B2Sample.java
when pointed at staging where Application Key API's are turned on.

**ApplicationKeys are still not available in production.


certainmagic:
this is the work referenced in issue #42.

PR #48, opened 2018-07-12
New params for b2_list_buckets and b2_list_unfinished_large_files

bwbeach:
The updated APIs just went live at Backblaze, so now it's time to put support into the SDK.


certainmagic:
those look fine to me.  please merge, etc.

PR #53, opened 2018-09-07
Fix #507: B2json versions

bwbeach:
Adds the ability to say that some fields of a JSON object exist in only some versions.

### DIFF
--- a/core/src/main/java/com/backblaze/b2/json/B2Json.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2Json.java
@@ -285,6 +285,9 @@ public class B2Json {
         return fromJsonUntilEof(in, clazz, B2JsonOptions.DEFAULT);
     }
 
+    /**
+     * Use the call that takes B2JsonOptions, no this one with 'int optionFlags'.
+     */
     @Deprecated
     public <T> T fromJsonUntilEof(InputStream in, Class<T> clazz, int optionFlags) throws IOException, B2JsonException {
         return fromJsonUntilEof(in, clazz, optionsFromFlags(optionFlags));
@@ -308,6 +311,9 @@ public class B2Json {
         return fromJson(in, clazz, B2JsonOptions.DEFAULT);
     }
 
+    /**
+     * Use the call that takes B2JsonOptions, no this one with 'int optionFlags'.
+     */
     @Deprecated
     public <T> T fromJson(InputStream in, Class<T> clazz, int optionFlags) throws IOException, B2JsonException {
         return fromJson(in, clazz, optionsFromFlags(optionFlags));
@@ -332,6 +338,9 @@ public class B2Json {
         return fromJson(json, clazz, B2JsonOptions.DEFAULT);
     }
 
+    /**
+     * Use the call that takes B2JsonOptions, no this one with 'int optionFlags'.
+     */
     @Deprecated
     public <T> T fromJson(String json, Class<T> clazz, int optionFlags) throws B2JsonException {
         return fromJson(json, clazz, optionsFromFlags(optionFlags));
@@ -342,6 +351,9 @@ public class B2Json {
         return fromJsonWithHandler(json, handler, options);
     }
 
+    /**
+     * Use the call that takes B2JsonOptions, no this one with 'int optionFlags'.
+     */
     @Deprecated
     private <T> T fromJsonWithHandler(String json, B2JsonTypeHandler handler, int optionFlags) throws B2JsonException {
         return fromJsonWithHandler(json, handler, optionsFromFlags(optionFlags));
@@ -364,6 +376,9 @@ public class B2Json {
         return fromJson(jsonUtf8Bytes, clazz, B2JsonOptions.DEFAULT);
     }
 
+    /**
+     * Use the call that takes B2JsonOptions, no this one with 'int optionFlags'.
+     */
     @Deprecated
     public <T> T fromJson(byte[] jsonUtf8Bytes, Class<T> clazz, int optionFlags) throws IOException, B2JsonException {
         return fromJson(jsonUtf8Bytes, clazz, optionsFromFlags(optionFlags));
@@ -387,6 +402,9 @@ public class B2Json {
         return fromUrlParameterMap(parameterMap, clazz, B2JsonOptions.DEFAULT);
     }
 
+    /**
+     * Use the call that takes B2JsonOptions, no this one with 'int optionFlags'.
+     */
     @Deprecated
     public <T> T fromUrlParameterMap(Map<String, String> parameterMap, Class<T> clazz, int optionFlags) throws IOException, B2JsonException {
         return fromUrlParameterMap(parameterMap, clazz, optionsFromFlags(optionFlags));
--- a/core/src/main/java/com/backblaze/b2/json/B2Json.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2Json.java
@@ -424,6 +424,25 @@ public class B2Json {
     public @interface ignored {}
 
     /**
+     * Annotation that says that a field exists in all versions at or after this one.
+     */
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.FIELD)
+    public @interface firstVersion {
+        int firstVersion();
+    }
+
+    /**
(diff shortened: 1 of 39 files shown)


core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -1334,6 +1334,42 @@ public void testOptionalWithDefault() throws B2JsonException {
...
+
+        @B2Json.constructor(params =  "n")
+        private VersionRangeBackwardsClass(int n) {
+            this.n = n;
+        }
+    }
+
+    @Test
+    public void testConflictingVersions() throws B2JsonException {
+        thrown.expectMessage("must not specify both 'firstVersion' and 'versionRange'");
```
certainmagic: it might be more dev-friendly if this message included the name of the class with the issue.

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -1801,4 +1837,77 @@ private SubclassH(String name) {
...
+        assertEquals(0, obj.x);
+        assertEquals(1, obj.version);
+    }
+
+    @Test
+    public void testRequiredFieldMissingInVersion() throws B2JsonException {
+        final String json = "{}";
+        final B2JsonOptions options = B2JsonOptions.builder().setVersion(5).build();
+
+        thrown.expectMessage("required field x is missing");
```
certainmagic: can we say which class has the problem?
please, pretend i'm asking this for all new error messages. :)

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -1801,4 +1837,77 @@ private SubclassH(String name) {
...
+        final String json = "{ \"x\": 5 }";
+        final B2JsonOptions options = B2JsonOptions.builder().setVersion(10).build();
+        final VersionedContainer obj = bzJson.fromJson(json, VersionedContainer.class, options);
+        assertEquals(5, obj.x);
+        assertEquals(10, obj.version);
+    }
+
+    @Test
+    public void testSerializeSkipFieldNotInVersion() throws B2JsonException {
+        final B2JsonOptions options = B2JsonOptions.builder().setVersion(1).build();
```
certainmagic: would setVersion(3) be better?  :)

certainmagic: sorry if i just haven't seen it yet or if it hasn't sunk in.  are there any tests for a field going away (to test lastVersion?).

core/src/main/java/com/backblaze/b2/json/B2Json.java
```diff
@@ -245,10 +282,15 @@ public String listToJson(List<?> list, Class<?> valueClass) throws B2JsonExcepti
      * if there is anything but whitespace after the JSON value.
      */
     public <T> T fromJsonUntilEof(InputStream in, Class<T> clazz) throws IOException, B2JsonException {
-        return fromJsonUntilEof(in, clazz, 0);
+        return fromJsonUntilEof(in, clazz, B2JsonOptions.DEFAULT);
+    }
+
+    @Deprecated
```
certainmagic: willing to add a comment to @Deprecateds which say what to use instead? even if it's generic, like "use the B2JsonOptions version instead"?

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -138,32 +144,43 @@
...
 
-            int bitMask = 0;
+            Integer versionParamIndex = null;
             for (int i = 0; i < paramNames.length; i++) {
                 String paramName = paramNames[i];
-                final FieldInfo fieldInfo = fieldMap.get(paramName);
-                if (fieldInfo == null) {
-                    throw new B2JsonException(clazz.getName() + " param name is not a field: " + paramName);
+                if (paramName.isEmpty()) {
+                    throw new B2JsonException("parameter name must not be empty");
```
certainmagic: include class name in error message?

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -138,32 +144,43 @@
...
+                }
+                if (paramName.equals(versionParamOrEmpty)) {
+                    versionParamIndex = i;
```
certainmagic: check that we don't have versionParamIndex yet?

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -315,13 +356,15 @@ public void serialize(T obj, B2JsonWriter out) throws IOException, B2JsonExcepti
...
                         typeFieldDone = true;
                     }
-                    out.writeObjectFieldNameAndColon(fieldInfo.getName());
-                    final Object value = fieldInfo.field.get(obj);
-                    if (fieldInfo.requirement == FieldRequirement.REQUIRED && value == null) {
-                        throw new B2JsonException("required field " + fieldInfo.getName() + " cannot be null");
+                    if (fieldInfo.isInVersion(version)) {
+                        out.writeObjectFieldNameAndColon(fieldInfo.getName());
+                        final Object value = fieldInfo.field.get(obj);
+                        if (fieldInfo.isRequiredInVersion(version) && value == null) {
```
certainmagic: the code here and the existence of isRequiredInVersion makes me think that i should be able to say "became option in version X" or "required starting in version Y", but i don't know how to do that. right now, i *think* requiredness is orthogonal to version info, so it's not yet possible to express those more complicated restrictions.  that's fine -- as long as we don't need them, simpler is better.  :)

i just want to check....am i missing something?

certainmagic: maybe isRequiredInVersion() is really isRequiredAndInVersion() right now?  (it's implemented that way.)

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -423,48 +478,48 @@ public T deserializeFromUrlParameterMap(Map<String, String> parameterMap, int op
...
                 if (fieldInfo.defaultValueOrNull != null) {
                     constructorArgs[index] = fieldInfo.defaultValueOrNull;
                 }
                 else {
                     constructorArgs[index] = fieldInfo.handler.defaultValueForOptional();
                 }
             }
+            else {
+                if (!fieldInfo.isInVersion(version)) {
+                    throw new B2JsonException("field " + fieldInfo.getName() + " is not in version " + version);
```
certainmagic: is there a test for this?

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
bwbeach: done.

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
bwbeach: This is not a new error message, and...

B2Json errors show through the public B2 APIs at Backblaze.  When a client passes in JSON with a missing field, this error message is returned.  The Java class name won't make sense to the caller.

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
bwbeach: ok

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
bwbeach: will add

core/src/main/java/com/backblaze/b2/json/B2Json.java
bwbeach: ok

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
bwbeach: ok

PR #54, opened 2018-09-11
Switch to using B2 API version 2

bwbeach:
I think this change implies a new major version number, because the behavior of the calls will be changing.


core/src/main/java/com/backblaze/b2/client/structures/B2FileVersion.java
```diff
@@ -10,6 +10,14 @@
 import java.util.Map;
 import java.util.Objects;
 
+/**
+ * Information about a file, as returned by the B2 API.
+ *
+ * The API returns two fields that are not included here: accountId and bucketId.
+ * The reason for not including them is that this SDK also returns the same
+ * structure from getFileInfoByName, which gets the info from the headers returned
+ * by a HEAD request on the file, which do no include the
```
certainmagic: missing "m" on last word? :)

PR #61, opened 2019-03-13
B2 json from json or throw

spartan218:
Reminder that bzJson variable in B2JsonTest was renamed to b2Json.


spartan218:
Not sure why the version file was changed marked as changed, nothing changed.

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -1943,7 +1943,78 @@ public VersionedContainer(int x, int v) {
...
+    @Test
+    public void testToJson() {
+        final TestClassOne testClassOne = new TestClassOne("testABC", 1);
+        final String testClassOneStr = B2Json.toJsonOrThrowRuntime(testClassOne);
+        assertEquals("{\n  \"name\": \"testABC\",\n  \"number\": 1\n}", testClassOneStr);
+    }
+
+    @Test
+    public void testToJsonWithOptions() {
+        final B2JsonOptions options = B2JsonOptions.builder().build();
```
certainmagic: might be useful to test with a non-default option.

spartan218:
@certainmagic added a test with a version number difference in the options, let me know if thats what you were thinking.

certainmagic:
yep. thanks, briang.  :)

my goal was to just have something that showed the options were passed through in the one that took both arguments. i don't think we need both a default-options and non-default-options test.   but this is fine.  :)

PR #62, opened 2019-03-18
app key renaming

spartan218:
Places where we need to rename accountID to application key ID moving forward.

Note: that account ID still persists in certain areas as it is needed for other b2 operations.


core/src/main/java/com/backblaze/b2/client/B2AccountAuthorizerSimpleImpl.java
```diff
@@ -14,43 +14,43 @@
  * Almost everyone is expected to use this right now.
  */
 public class B2AccountAuthorizerSimpleImpl implements B2AccountAuthorizer {
-    private final String accountId;
+    private final String applicationKeyId;
```
certainmagic: the claim is that changing this does NOT change the API nor does it change what we send on the wire, right?

core/src/main/java/com/backblaze/b2/client/credentialsSources/B2Credentials.java
```diff
@@ -9,9 +9,9 @@
  */
 public interface B2Credentials {
     /**
-     * @return the accountId to use for b2_authorize_account.
+     * @return the applicationKeyId to use for b2_authorize_account.
      */
-    String getAccountId();
+    String getApplicationKeyId();
```
certainmagic: this is an incompatible API change, methinks.  with this change, i think we will need to bump the major version number of the SDK.  is that necessary?

core/src/main/java/com/backblaze/b2/client/credentialsSources/B2CredentialsFromEnvironmentSource.java
```diff
@@ -8,7 +8,7 @@
 
 /**
  * This B2CredentialsSource reads the credentials from two environment variables:
- *   B2_ACCOUNT_ID
+ *   B2_APPLICATION_KEY_ID
```
certainmagic: i think this change means that people who run programs which use this class will have their programs stop working until they change their environment variables.  is that necessary?

core/src/main/java/com/backblaze/b2/client/credentialsSources/B2Credentials.java
spartan218: Can you help me understand why this is an incompatible API change?

core/src/main/java/com/backblaze/b2/client/credentialsSources/B2CredentialsFromEnvironmentSource.java
spartan218: If they upgrade to the newest SDK once this is merged I believe that they need to change their environment variables names, which is documented.

 I'm not sure how to answer your question about being necessary. The goal behind the app key renaming is to bring clarity to the name. My intention with this change was to bring b2-java-sdk users up to speed on the same terminology that we want everyone to use.

certainmagic: briang & i talked and he's taking a stab at maintaining compatibility.

README.md
```diff
@@ -54,7 +54,7 @@ SAMPLE
   * To run B2Sample, you will need to add your credentials to your environment
     in these environment variables:
 
-    * B2_APPLICATION_KEY_ID
+    * B2_ACCOUNT_ID OR B2_APPLICATION_KEY_ID
```
certainmagic: thanks.  :)

core/src/main/java/com/backblaze/b2/client/credentialsSources/B2CredentialsImpl.java
```diff
@@ -9,19 +9,30 @@
  */
 public class B2CredentialsImpl implements B2Credentials {
     private final String accountId;
+    private final String appKeyId;
     private final String appKey;
 
     B2CredentialsImpl(String accountId,
                              String appKey) {
+
+        // for now consider any accountId an appKeyId
```
certainmagic: that *sounds* right.  let's talk it through together.

core/src/main/java/com/backblaze/b2/client/structures/B2AuthorizeAccountRequest.java
```diff
@@ -10,25 +10,25 @@
...
+    public static Builder builder(String applicationKeyId,
                                   String applicationKey) {
-        return new Builder(accountId, applicationKey);
+        return new Builder(applicationKeyId, applicationKey);
 
     }
 
-    public String getAccountId() {
-        return accountId;
+    public String getApplicationKeyId() {
```
certainmagic: looks like we're losing getAccountId() here too.
this whole trying-to-be-compatible is awkward.

spartan218: yea I can add it back.

PR #68, opened 2019-04-29
Okhttp2

OldSkoolMark:
Provide Android-favored alternative to Apache HTTP Client that uses OkHttp. OkHttp is pure Java, so Android builds should just need to reference core and okhttp maven artifacts.


okhttp/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+    pomDescription = 'OkHttp support for B2 SDK for Java.'
+}
+
+apply from: '../common.gradle'
+
+dependencies {
+    implementation 'com.backblaze.b2:b2-sdk-core:2.0.0'
+    implementation "com.squareup.okhttp3:okhttp:3.12.1"
+    implementation "com.squareup.okhttp3:okhttp-urlconnection:3.12.1"
+    implementation 'org.jetbrains:annotations:15.0'
```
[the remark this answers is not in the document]
certainmagic: or is okhttp using it?

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
```diff
@@ -0,0 +1,424 @@
...
+            return str.getBytes(UTF8);
+        } catch (UnsupportedEncodingException e) {
+            // this is very, very bad and it's not gonna get better by itself.
+            throw new RuntimeException("No UTF-8 charset", e);
+        }
+    }
+
+
+
+    private final static String TAG = B2OkHttpClientImpl.class.getSimpleName();
```
certainmagic: what's using this?

check_code
```diff
@@ -52,6 +52,7 @@ NUMBER_OF_LINES_TO_EXAMINE = 10
 LAYERS = [
     'sample',
     'client.webApiHttpClient',
+    'client.okHttpClient'
```
certainmagic: i think this is missing a comma.
and i think that's what's causing the build to fail with new messages.

okhttp/build.gradle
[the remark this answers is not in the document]
OldSkoolMark: @Nullable etc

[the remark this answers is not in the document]
certainmagic: it looks like maybe we're using @NotNull and @Nullable.  these methods are only being used in this one file and, since they're not adding value to b2-sdk-java users, i'd like to minimize our external dependencies.

how would you feel about removing this dependency?

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
OldSkoolMark: Android world

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    private Response maybeDecorateResponse( Response originalResponse ){
+        if( progressListener == null){
+            return originalResponse;
+        } else {
+            Request request = originalResponse.request();
+            if (request.method().equalsIgnoreCase("GET")
+                    && request.url().toString().contains("b2_download_file")) {
+                String b2FileID = request.url().queryParameter("fileId");
```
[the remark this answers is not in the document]
OldSkoolMark: If it's in the headers we can get it. My Android client doesn't need download by name (yet), but progress reporting should be available for all of the download options I agree. Part of the discussion ...

okhttp/build.gradle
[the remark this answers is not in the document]
OldSkoolMark: removed

core/src/main/java/com/backblaze/b2/client/credentialsSources/B2CredentialsFromEnvironmentSource.java
```diff
@@ -76,7 +76,7 @@ private String getApplicationKeyIdOrThrow() throws IllegalArgumentException{
             // maybe fallback to using the old variable
             final String accountId = System.getenv("B2_ACCOUNT_ID");
             if (accountId != null) {
-                B2Preconditions.checkState( accountId.isEmpty(),
+                B2Preconditions.checkState( !accountId.isEmpty(),
```
certainmagic: good catch.

spartan218: Can you pull this out of this PR and make a new one with just this change?

spartan218: nvm ill do it

certainmagic: @spartan218 just merged this fix into master because it's important.  :)

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
```diff
@@ -0,0 +1,317 @@
...
+     * @param headersOrNull the headers, if any.
+     *
+     * @return the body of the response.
+     * @throws B2Exception if there's any trouble
+     */
+    private  String postAndReturnString( String url,  B2Headers headersOrNull,  InputStream inputStream, long contentLength)
+            throws B2Exception {
+        byte[] bytes;
+        try {
+            bytes = readFully(inputStream, (int)contentLength);
```
certainmagic: maybe use B2Preconditions to make sure that contentLength is small enough to not lose data with this cast?

OldSkoolMark: What should I do if it is too large?

OldSkoolMark: As discussed, I now use a temporary file when there are too many bytes for a byte[]

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
```diff
@@ -0,0 +1,327 @@
...
+        // Closing response body streams is handled elsewhere by OkHttp.
+    }
+
+    private String postJsonAndReturnString(String url,
+                                           B2Headers headersOrNull,
+                                           Object request) throws B2Exception {
+        try {
+            B2Json bzJson = B2Json.get();
+            String requestJson = bzJson.toJson(request);
+            byte[] requestBytes = getUtf8Bytes(requestJson);
```
certainmagic: hmmm....hidden limitation that we can't handle more than 2GB-ish of json in our requests.  that's the same thing exists in the Apache HttpClient-based implementation.  i think that's fine -- B2 request json structures are relatively small.

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
```diff
@@ -0,0 +1,327 @@
...
+    private  String postAndReturnString( String url,  B2Headers headersOrNull,  InputStream inputStream, long contentLength)
+            throws B2Exception {
+        RequestBody body;
+        try {
+            if( contentLength <= (Integer.MAX_VALUE-HOTSPOT_FUDGE) ){
+                byte[] bytes = readFully(inputStream, (int) contentLength);
+                body = RequestBody.create(MediaType.get(APPLICATION_JSON), bytes);
+            } else {
+                String tempFileName = UUID.randomUUID().toString();
+                File tempFile = File.createTempFile(tempFileName, ".tmp");
```
certainmagic: i think we should call tempFile.deleteOnExit() immediately after this line so that we don't leak it if we fail part of the way through the copy() below.

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
```diff
@@ -0,0 +1,327 @@
...
+                FileOutputStream outputStream = new FileOutputStream(tempFile);
+                copy( inputStream, outputStream);
```
certainmagic: copy() doesn't close the destination stream.  i think you need to do that to make sure everything has made it to disk.

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
certainmagic: it also seems like we should hold on to this temp file name and delete it in a finally block that covers the whole method.  we don't really want to wait until the app exits to delete the file.

certainmagic: the documentation[1] says that Files.createTempFile() "may be more suited to security-sensitive applications".  (note that's on Files, not File).  should we be using that instead?

[1] https://developer.android.com/reference/java/io/File#createTempFile(java.lang.String,%2520java.lang.String,%2520java.io.File)

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2StorageOkHttpClientBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+
+public class B2StorageOkHttpClientBuilder {
+
+    private static final String DEFAULT_MASTER_URL = "https://api.backblazeb2.com/";
+
+    private final B2ClientConfig config;
+    private Supplier<B2RetryPolicy> retryPolicySupplier;
+
+    @SuppressWarnings("WeakerAccess")
+    public static B2StorageOkHttpClientBuilder builder(String userAgent)  {
```
certainmagic: now that i have added B2StorageClientFactory and it has a method that does this, perhaps you don't need it here?  I don't think I can remove it from the Apache HttpClient implementation because we've been shipping that way for years, but maybe we don't need to add it here?

What do you think?  Let's see what you come up with after looking at the new B2StorageClientFactory code.

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2StorageOkHttpClientBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+        return builder(credentials.getApplicationKeyId(), credentials.getApplicationKey(), userAgent );
+    }
+
+    @SuppressWarnings("WeakerAccess")
+    public static B2StorageOkHttpClientBuilder builder(B2ClientConfig config) {
+        return new B2StorageOkHttpClientBuilder(config);
+    }
+
+    @SuppressWarnings("WeakerAccess")
+    public static B2StorageOkHttpClientBuilder builder(String applicationKeyID, String applicationKey, String userAgent) {
```
certainmagic: same comment as above about whether this signature is needed.

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
spartan218: Is there a way to open the stream using a try with resources that way we don't need to remember to close it?
Just a thought, might not be a solution.

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
```diff
@@ -221,20 +223,25 @@ private String postJsonAndReturnString(String url,
...
+                Path tempFilePath = Files.createTempFile(tempFileName, ".tmp");
+                tempFile = tempFilePath.toFile();
                 FileOutputStream outputStream = new FileOutputStream(tempFile);
                 copy( inputStream, outputStream);
                 body = RequestBody.create(MediaType.get(APPLICATION_JSON), tempFile);
-                tempFile.deleteOnExit();
             }
         } catch (IOException e) {
             throw translateToB2Exception(e, url);
+        } finally {
```
certainmagic: i'm concerned that this delete is too early.
doesn't the file need to still exist during the okHttpClient.newCall() below?

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2StorageOkHttpClientBuilder.java
certainmagic: what do you think, @OldSkoolMark.  should we ditch this and the one i marked below?

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
certainmagic: @OldSkoolMark -- do you need to close the stream here?

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
certainmagic: hey @OldSkoolMark --

you addressed my second and third comment in this area, but i'm not sure you saw my first one...

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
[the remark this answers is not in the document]
certainmagic: gonna mark this thread as resolved since the progress stuff will be in a later PR.

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
certainmagic: how are you testing the file-is-too-big-for-an-array code path?

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
OldSkoolMark: I refactored this part of the code a bit. I think your concerns are now addressed.

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
OldSkoolMark: now closing immediately after the copy

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2OkHttpClientImpl.java
OldSkoolMark: fixed

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2StorageOkHttpClientBuilder.java
OldSkoolMark: If http-client has it I think we okhttp should too. That makes it easier for any client code that uses  those static methods to switch to the other http client.

okhttp/src/main/java/com/backblaze/b2/client/okHttpClient/B2StorageOkHttpClientBuilder.java
OldSkoolMark: same reply as above.

okhttp/src/test/java/com/backblaze/b2/client/okHttpClient/B2StorageHttpClientFactoryTest.java
```diff
@@ -0,0 +1,30 @@
...
+package com.backblaze.b2.client.okHttpClient;
+
+import com.backblaze.b2.client.B2StorageClientFactory;
+import com.backblaze.b2.client.B2StorageClientFactoryPathBasedImpl;
+import org.junit.Test;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+public class B2StorageHttpClientFactoryTest {
```
certainmagic: should this be  B2StorageOkHttpClientFactoryTest  (not the "ok" in the middle of the name)?

PR #70, opened 2019-05-02
Adding the sensitive annotation

malaysf:
Adding @B2Json.sensitive annotation to fields to allow them to be redacted when outputting Json with the setRedactSensitive option on. The output is still valid JSON but no longer will conform to the structure specified, so this cannot be used to round-trip the JSON, however is useful when outputting structures to log files where certain information should not be exposed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ core/build/
 httpclient/build/
 samples/build/
 b2-sdk-java.iml
+out/
+core/out/
--- a/core/src/main/java/com/backblaze/b2/json/B2JsonOptions.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2JsonOptions.java
@@ -43,11 +43,22 @@ public class B2JsonOptions {
     private final int version;
 
     /**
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -5,3 +5,5 @@ core/build/
 httpclient/build/
 samples/build/
 b2-sdk-java.iml
+out/
+core/out/
```
bwbeach: I think that red thing means a missing newline

core/src/main/java/com/backblaze/b2/json/B2JsonOptions.java
```diff
@@ -42,12 +42,23 @@
      */
     private final int version;
 
+    /**
+     * Whether to redact sensitive fields
+     *
+     * When set, fields marked as @B2Json.sensitive will be redacted from the serialized
+     * JSON and replaced with the string value "***REDACTED***"
+     * The output will be valid Json but the structure/types will not conform to the expected
```
bwbeach: picky spelling: s/Json/JSON/

PR #72, opened 2019-05-15
add B2StorageClientFactory: a way to create a B2StorageClient without…

certainmagic:
… directly naming an implementation.

we're working on adding an OkHttp-based implementation of B2WebApiClient
and we don't want to have the sample programs have to import a specific
web client implementation.  instead, they can now just say:

   B2StorageClientFactory.createDefaultFactory().create(USER_AGENT)

to create a B2StorageClient with whichever of the known implementations
are in the classpath.  in the future, we can allow people to register
their own classes, if that makes sense.

testing:
  * ./gradlew build javadoc writeNewPom
  * ran the unit tests in IDEA, including with coverage to check new tests.
  * ran B2 and B2Sample to make sure clients were created and usable.
  * Mark will uncomment (and maybe tweak) the OkHttp class registration
    in B2StorageClientFactoryPathBasedImpl then make sure that the sample
    programs work with his new implementation without having to change
    the sample code.

### DIFF
--- a/core/src/main/java/com/backblaze/b2/client/B2StorageClientFactory.java
+++ b/core/src/main/java/com/backblaze/b2/client/B2StorageClientFactory.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2019, Backblaze Inc. All Rights Reserved.
+ * License https://www.backblaze.com/using_b2_code.html
+ */
+package com.backblaze.b2.client;
+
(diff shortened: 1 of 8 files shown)


certainmagic:
with approval here from @bwbeach and a nod from mark over slack, i'm gonna merge this.  :)

PR #76, opened 2019-06-12
Implement progress reporting in the B2LargeFileStorer

generictjohnson:


core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
```diff
@@ -158,39 +162,138 @@ private B2FileVersion finishLargeFileFromB2PartFutures(B2FileVersion largeFileVe
...
-                accountAuthCache, () -> webifier.copyPart(accountAuthCache.get(), copyPartRequest),
-                retryPolicySupplier.get());
+        try {
+            return retryer.doRetry(
+                    "b2_copy_part",
+                    accountAuthCache, () -> {
+                        updateProgress(
+                                uploadListener,
+                                partNumber,
+                                1,
```
certainmagic: comment?

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
```diff
@@ -158,39 +162,138 @@ private B2FileVersion finishLargeFileFromB2PartFutures(B2FileVersion largeFileVe
...
+            B2UploadState uploadState) {
 
-        return retryer.doRetry(
-                "b2_upload_part",
-                accountAuthCache,
-                (isRetry) -> {
-                    final B2UploadPartUrlResponse uploadPartUrlResponse = uploadPartUrlCache.get(isRetry);
+        uploadListener.progress(
+                new B2UploadProgress(
+                        partNumber - 1, partStorers.size(), -1, partLength, bytesSoFar, uploadState));
```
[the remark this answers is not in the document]
certainmagic: we should probably also call out in the docs for B2UploadProgress that startByte can be -1, too!
that's probably surprising.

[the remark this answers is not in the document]
certainmagic: ideally, when we *can* know the startByte, it'd be nice to, but that's probably more effort than it's worth?  it seems like something the code that has a list of all the part storers could figure out and tell each storer, when it's known.  if/when we switch to having normal large file uploads go through here, it's probably a requirement so we don't lose information we've historically given.

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
```diff
@@ -158,39 +162,138 @@ private B2FileVersion finishLargeFileFromB2PartFutures(B2FileVersion largeFileVe
...
                 .build();
 
-        return retryer.doRetry(
-                "b2_copy_part",
-                accountAuthCache, () -> webifier.copyPart(accountAuthCache.get(), copyPartRequest),
-                retryPolicySupplier.get());
+        try {
+            return retryer.doRetry(
+                    "b2_copy_part",
+                    accountAuthCache, () -> {
```
certainmagic: nitpick?  don't start the lambda on the same line as the previous argument?

core/src/main/java/com/backblaze/b2/client/B2StorageClient.java
```diff
@@ -237,6 +238,7 @@ B2FileVersion uploadLargeFile(B2UploadFileRequest request,
      * @param fileVersion The B2FileVersion for the large file getting stored.
      *                    This is the return value of startLargeFile().
      * @param contentSource The contentSource to upload.
+     * @param uploadListener The object that handles upload progress events.
```
certainmagic: can you look at other places we let you provide upload listeners?  does this follow our style?  in particular, should we allow null (and convert it to a noopListener internally?)

core/src/test/java/com/backblaze/b2/client/B2LargeFileStorerTest.java
```diff
@@ -252,4 +308,39 @@ public boolean equals(Object o) {
         }
     }
 
+    /**
+     * An executor that runs tasks in the order they are submitted. It accomplishes this by running them in the main
+     * thread. This is done so our assertions can assume a specific order of progress events. This works because the
+     * tasks are all independent of each other and the main thread has no work to do while the tasks are running.
+     */
+    private class ExecutorThatUsesMainThread extends AbstractExecutorService {
```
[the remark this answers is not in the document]
certainmagic: and...Tim just pointed out to me that this would be the first use of guava, and we definitely don't want to add a dependency.   thx, Tim.  :)

core/src/main/java/com/backblaze/b2/client/B2StorageClient.java
generictjohnson: Providing that API popped into my head. Since you suggested it too I can take care of that. For uploading a large file, the upload listener is provided via the `B2UploadFileRequest` which is passed to the storage client. I figured this was the closest thing. Your thoughts?

certainmagic: B2UploadFileRequest's constructor says this:
        this.listener = (listener != null) ? listener : B2UploadListener.noopListener();

and the builder for that structure defaults to null, so i'm inclined to let people pass null in your case.

and this discussion makes me think about the fact that many of the methods on the storage client take a request structure that is build with a builder.  this lets us add new options without causing binary or source-level incompatibilities.

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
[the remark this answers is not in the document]
generictjohnson: My next push will reflect the following changes:

Added a constant to B2UploadProgress for unknown start byte, as well as a method isStartByteKnown();
Added a NOTE in the top-level javadoc for B2UploadProgress.
Added code to compute the starting bytes for all parts before the first copied part.

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
generictjohnson: I added a constant `UNKNOWN_PART_SIZE_PLACEHOLDER` which hopefully clarifies things.

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
generictjohnson: Done.

core/src/test/java/com/backblaze/b2/client/B2LargeFileStorerTest.java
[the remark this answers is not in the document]
generictjohnson: np :)

core/src/main/java/com/backblaze/b2/client/B2StorageClient.java
generictjohnson: Are you suggesting creating a StoreLargeFile request with a corresponding builder?

certainmagic:
Yes and no.  :)

That's what I'm referencing, but i think a reasonable idea is that if we
ended up with more options, we make a request structure later and make this
version a convenience method for calling it.  :)

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
```diff
@@ -77,6 +86,28 @@
         return partStorers;
     }
 
+    private static List<Long> computeStartingBytePositions(List<B2PartStorer> partStorers) {
+        final List<Long> startingPositions = new ArrayList<>();
```
certainmagic: maybe provide the initial capacity?

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
```diff
@@ -77,6 +86,28 @@
...
+        } catch (B2CannotComputeException e) {
+            while (startingPositions.size() < partStorers.size()) {
+                startingPositions.add(B2UploadProgress.UNKNOWN_PART_START_BYTE);
+            }
+        }
+
+        return startingPositions;
+    }
+
+    long getStartByte(int partNumber) {
```
certainmagic: "...OrUnknown"? or maybe just a comment to point out that it might return UNKNOWN_PART_START_BYTE?

core/src/main/java/com/backblaze/b2/client/structures/B2UploadProgress.java
```diff
@@ -14,8 +14,35 @@
  *       For example, if we're adding a SHA1 to the thing being uploaded,
  *       the contentLength will be at least 40 bytes longer than you expect
  *       because we're adding an ascii-hex-encoded SHA1.
+ *
+ * NOTE: when storing a large file that involves copying parts, it no
```
certainmagic: "it no longer becomes possible"?  "it's not possible"?

core/src/main/java/com/backblaze/b2/client/structures/B2UploadProgress.java
```diff
@@ -14,8 +14,35 @@
...
+ *       parts will show the length as UNKNOWN_PART_SIZE_PLACEHOLDER. Once a copy succeeds,
+ *       the progress event for that success will contain the correct length.
  */
 public class B2UploadProgress {
+
+    /**
+     * Placeholder value for part sizes that will not be known until the
+     * part is stored in B2. For copy operations, we will not always know
+     * the true part size until we receive the B2Part from b2_copy_part.
+     */
```
certainmagic: noe that since no part can be empty, this is always right or an underestimate?  :)

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
```diff
@@ -77,6 +86,28 @@
         return partStorers;
     }
 
+    private static List<Long> computeStartingBytePositions(List<B2PartStorer> partStorers) {
+        final List<Long> startingPositions = new ArrayList<>();
+
+        long cursor = 0;
+        try {
+            for (final B2PartStorer partStorer : partStorers) {
```
certainmagic: one of your test cases has the same partNumber more than once.  that makes me wonder what we're assuming about part number ordering in this list...  what happens with unsorted lists? what happens with duplicate part numbers?

core/src/test/java/com/backblaze/b2/client/B2LargeFileStorerTest.java
```diff
@@ -106,6 +105,49 @@ private B2LargeFileStorer createFromLocalContent() throws B2Exception {
...
+        when(contentSource.getContentLength()).thenReturn(size);
+        return contentSource;
+    }
+
+    private B2LargeFileStorer createLargeFileStorerForStartByteTests() throws IOException {
+        final List<B2PartStorer> partStorers = Arrays.asList(
+                new B2UploadingPartStorer(1, createContentSourceWithSize(100)),
+                new B2AlreadyStoredPartStorer(part2),
+                new B2CopyingPartStorer(3, fileId(3)),
+                new B2UploadingPartStorer(1, createContentSourceWithSize(900)));
```
certainmagic: duplicate part number?  is that intentional?  see questions above.  :)

core/src/test/java/com/backblaze/b2/client/B2LargeFileStorerTest.java
```diff
@@ -294,7 +354,7 @@ public Long getSrcLastModifiedMillisOrNull() {
         @Override
         public InputStream createInputStream() {
             // will not get called
-            throw new NotImplementedException();
+            return null;
```
certainmagic: the comment says "will not get called".  since you changed it to not throw, i'm guesing that might not be true any more...?

generictjohnson: No there were some warnings about throwing that particular exception when I ran `gradlew test (or build- can't remember)` so I just changed it.

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
generictjohnson: Done.

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
generictjohnson: Done.

core/src/main/java/com/backblaze/b2/client/structures/B2UploadProgress.java
generictjohnson: Done.

core/src/main/java/com/backblaze/b2/client/structures/B2UploadProgress.java
generictjohnson: Technically a part cannot be less than 5MB, so I strictly said it could be an estimate :)

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
generictjohnson: I wrote some validation logic to sort and detect duplicates and gaps. I argue that a gap is a problem too because we need all of the SHA1s to finish the file.

core/src/test/java/com/backblaze/b2/client/B2LargeFileStorerTest.java
generictjohnson: Good catch! This is now being addressed with the validation logic I added :)

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
```diff
@@ -158,39 +227,156 @@ private B2FileVersion finishLargeFileFromB2PartFutures(B2FileVersion largeFileVe
...
+                    retryPolicySupplier.get());
+        } catch (Exception e) {
+            updateProgress(
+                    uploadListener,
+                    partNumber,
+                    B2UploadProgress.UNKNOWN_PART_SIZE_PLACEHOLDER,
+                    0,
+                    B2UploadState.FAILED);
+
+            throw e;
```
bwbeach: How can you throw `Exception` here when the method throws the more specific `B2Exception`?

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
```diff
@@ -42,6 +48,13 @@
      */
     private final List<B2PartStorer> partStorers;
 
+    /**
+     * Stores where each part will begin in the finished large file.
+     * Will be set to B2UploadProgress.UNKNOWN_PART_START_BYTE for
+     * parts that come after a copied part.
+     */
+    private final List<Long> startingBytePositions;
```
bwbeach: It's possible for them all to be UNKNOWN_PART_START_BYTE.  What's the benefit of having these if you can't count on them?

core/src/main/java/com/backblaze/b2/client/structures/B2UploadProgress.java
bwbeach: The last part is allowed to be 1 byte.

generictjohnson: Good point, I forgot about that.

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
generictjohnson: ab and I discussed having the `B2LargeFileStorer` replace the `B2LargeFileUploader`. Since we didn't want to lose that functionality, we provide a best-effort solution. For purely uploaded files, the new behavior should be identical to the old.

core/src/main/java/com/backblaze/b2/client/B2LargeFileStorer.java
generictjohnson: Not sure why I was catching Exception - switched to B2Exception.

PR #77, opened 2019-06-20
b2_copy_file support

generictjohnson:


core/src/main/java/com/backblaze/b2/client/B2StorageClient.java
```diff
@@ -211,6 +212,16 @@ default B2Bucket getBucketOrNullByName(String name) throws B2Exception {
...
 
+    /**
+     * Makes a copy of a file in the same bucket.
+     * The new file must be smaller than the maximum file size (5 GB).
+     *
+     * @param request describes what file to copy, the range of bytes to copy, and how to handle file metadata.
+     * @return The B2FileVersion of the new file.
+     * @throws B2Exception if there's any trouble.
+     */
+    B2FileVersion copySmallFile(B2CopyFileRequest request) throws B2Exception;
```
bwbeach: Do we use the term "small file" elsewhere?  I'm not sure we have an official name for non-large files.

(The docs seem to use "normal file".)

core/src/main/java/com/backblaze/b2/client/structures/MetadataDirective.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright 2019, Backblaze Inc. All Rights Reserved.
+ * License https://www.backblaze.com/using_b2_code.html
+ */
+
+package com.backblaze.b2.client.structures;
+
+public enum MetadataDirective {
```
bwbeach: Are these ever returned by the service?  In other places, the SDK uses strings rather than enums to allow for new enum items defined by the service.

core/src/main/java/com/backblaze/b2/client/B2StorageClient.java
generictjohnson: This interface provides an `uploadSmallFile()` method, so I was basing off of that. Searching for "smallFile" in the code, it shows up a bunch, so I'm inclined to keep it as is, for consistency's sake.

core/src/main/java/com/backblaze/b2/client/structures/MetadataDirective.java
generictjohnson: They are not returned by the service. I can switch to strings to be safe.

PR #81, opened 2019-08-02
Split handler initialization into: constructor and initialize()

bwbeach:
This fixes the bug where you could not use union types that were
recursive.  There's now a test case for this.

There was some confusing code to handle recursive object types that
put handlers in the map halfway through their constructors.  Giving
access to an object before it's constructed is bad practice, and
confusing.  Now, there are two phases of initialization: the constructor
and an initialize() method.  And the object doesn't leak out during
the constructor.  The B2JsonInitializedTypeHandler javadoc describes
the rules about what happens when.

This makes it clear what you cannot look at other handlers in your
constructor, and what fields you're allowed to look at in other
handlers in your initialize() method.

Testing: all tests

### DIFF
--- a/core/src/main/java/com/backblaze/b2/json/B2JsonInitializedTypeHandler.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2JsonInitializedTypeHandler.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2019, Backblaze, Inc.  All rights reserved.
+ */
+
+package com.backblaze.b2.json;
+
+import com.backblaze.b2.util.B2Preconditions;
+
+/**
+ * Base class for all implementations of B2JsonTypeHandler that have an initialize() method.
+ *
+ * (De)serialization is expected to be fast, so each implementation class gathers
+ * information it needs when it's set up, so when it's time to run it has all of
+ * the necessary information at hand.  This gets tricky because dependencies between
+ * handlers may have loops.
+ *
+ * The plan is to initialize in two phases:
+ *
+ * First, the constructor, which must not depend on any other handlers, and which
+ * must gather all of the information that other handlers will need.
+ *
+ * Second, the initialize() method does any work that needs information from other
+ * type handlers.
+ *
+ * Both phases are protected by the lock on B2JsonHandlerMap, so they don't need to
+ * lock, and the data they store in the object is guaranteed to be visible without
+ * further locking.
+ *
+ * Methods that return data set during initialize() should include this check:
+ *
+ *     Preconditions.checkState(isInitialized());
+ */
+public abstract class B2JsonInitializedTypeHandler<T> implements B2JsonTypeHandler<T> {
+
+    /**
+     * Has the initialize() method been run?
+     */
+    private boolean initialized = false;
+
+    /**
+     * Does any setup that requires information from other handlers.
+     *
+     * This is package-private; we only expect this to be called from B2JsonHandlerMap
+     * while it holds its lock.
+     */
+    void initialize(B2JsonHandlerMap b2JsonHandlerMap) throws B2JsonException {
+        B2Preconditions.checkState(!initialized);
+        initializeImplementation(b2JsonHandlerMap);
+        initialized = true;
+    }
+
+    /**
+     * Does any initialization specific to the concrete class.
+     *
+     * Override in classes that need to gather information from other handlers.
+     */
+    protected void initializeImplementation(B2JsonHandlerMap b2JsonHandlerMap) throws B2JsonException {}
+
+    /**
+     * Has the initialize method run?
+     * @return
+     */
+    protected boolean isInitialized() {
+        return initialized;
+    }
+}
--- a/core/src/main/java/com/backblaze/b2/json/B2JsonInitializedTypeHandler.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2JsonInitializedTypeHandler.java
@@ -30,6 +30,9 @@ import com.backblaze.b2.util.B2Preconditions;
  * Methods that return data set during initialize() should include this check:
  *
  *     Preconditions.checkState(isInitialized());
+ *
+ * NOTE: adding the initialize() method to BzJsonTypeHandler would change the interface
+ * and break any clients who have written then own handlers.
  */
 public abstract class B2JsonInitializedTypeHandler<T> implements B2JsonTypeHandler<T> {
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/backblaze/b2/json/B2JsonInitializedTypeHandler.java
```diff
@@ -0,0 +1,67 @@
...
+ * information it needs when it's set up, so when it's time to run it has all of
+ * the necessary information at hand.  This gets tricky because dependencies between
+ * handlers may have loops.
+ *
+ * The plan is to initialize in two phases:
+ *
+ * First, the constructor, which must not depend on any other handlers, and which
+ * must gather all of the information that other handlers will need.
+ *
+ * Second, the initialize() method does any work that needs information from other
```
certainmagic: Since some code will be calling getUninitializedHandler() and warned to not count on things that aren't initialized yet, how can we communicate what that will be for type handlers the caller doesn't know about?  Is it more about which calls can't be called before initialize has been called?  And, does that leak into the B2JsonTypeHandler interface since serialize & deserialize are probably things that shouldn't be called until the class is initialized?

core/src/main/java/com/backblaze/b2/json/B2JsonUnionBaseHandler.java
```diff
@@ -93,7 +101,7 @@
             if (!hasSuperclass(typeClass, clazz)) { // use clazz.isAssignableFrom(typeClass)?
                 throw new B2JsonException(typeClass + " is not a subclass of " + clazz);
             }
-            final B2JsonTypeHandler<?> handler = handlerMap.getHandler(typeClass);
+            final B2JsonTypeHandler<?> handler = b2JsonHandlerMap.getHandler(typeClass);
```
certainmagic: why getHandler() instead of getUninitializedHandler()?  if there's a reason, maybe add a comment, since by default i would expect to only see getUninitializedHandler() calls during initializeImplementation().  ...or maybe i'm just confused.  :)

core/src/main/java/com/backblaze/b2/json/B2JsonInitializedTypeHandler.java
certainmagic: I'm wondering if we should just make initialize() a method on the base interface.

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -514,6 +521,9 @@ public T deserializeFromUrlParameterMap(Map<String, String> parameterMap, B2Json
     }
 
     private T deserializeFromConstructorArgs(Object[] constructorArgs, int version) throws B2JsonException {
+
+        B2Preconditions.checkState(isInitialized());
```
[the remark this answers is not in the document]
bwbeach: Carefully looking at all the methods.  I couldn't think of an automatic way to catch them all.

core/src/main/java/com/backblaze/b2/json/B2JsonInitializedTypeHandler.java
bwbeach: I'd be OK with moving `initialize()` up.  Would that make things clearer for you?

The rule is that constructors must assume that no handler they see is initialized.

core/src/main/java/com/backblaze/b2/json/B2JsonUnionBaseHandler.java
bwbeach: Good question.  Will fix.

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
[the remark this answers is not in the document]
bwbeach: I looked at each of the fields in the object, and flagged the methods that use fields not set in the constructor.

I couldn't think of a way to do it automatically.  :-(

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -309,12 +307,12 @@ private B2JsonTypeHandler getFieldHandler(Type fieldType, B2JsonHandlerMap handl
         if (fieldType instanceof Class) {
             final Class fieldClass = (Class) fieldType;
             //noinspection unchecked
-            return handlerMap.getHandler(fieldClass);
+            return handlerMap.getUninitializedHandler(fieldClass);
```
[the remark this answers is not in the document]
bwbeach: I'll got with `getUninitializedFieldHandler`

core/src/main/java/com/backblaze/b2/json/B2JsonInitializedTypeHandler.java
bwbeach: Oh.  Now I remember.  Moving the `initialize()` method up would change the interface, and break anybody who had their own implementation of it.  So I rejected that.

certainmagic: (brian and i talked separately.)

i was concerned that we were changing the meaning of the handler map in constructors and that these changes only give two-phase initialization to handlers that inherit from a particular abstract class.  i missed that we already don't pass the handler map to custom handlers.

so i'm ok leaving initialize as part of this subclass.

core/src/main/java/com/backblaze/b2/json/B2JsonInitializedTypeHandler.java
```diff
@@ -30,6 +30,9 @@
  * Methods that return data set during initialize() should include this check:
  *
  *     Preconditions.checkState(isInitialized());
+ *
+ * NOTE: adding the initialize() method to BzJsonTypeHandler would change the interface
+ * and break any clients who have written then own handlers.
```
certainmagic: "then own" --> "their own"?  :)

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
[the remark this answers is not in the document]
certainmagic: cool.

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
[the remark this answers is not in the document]
certainmagic: fair.

core/src/main/java/com/backblaze/b2/json/B2JsonUnionBaseHandler.java
certainmagic: thx.

Issue #46, opened 2018-07-03
New "union" support needs to support a "defaultForInvalid".


certainmagic:
We need a way to designate how to handle an unknown union type.  We want to be sure that clients running with an "old" idea of the allowed union subtypes can still parse json that contains a "new" union subtype.  The clients won't be able to process the new type, but they shouldn't get a parsing error.  Instead, when they look at the type in the field, it should be an instance of the type designated as being for unknown subtypes. 

This fills a similar compatibility need as the @B2Json.defaultForInvalidEnumValue annotation does for enums.

Here are some things I think need to be defined:

  * what do we call this thing?

  * how to mark the special class?
     * with an annotation on the type (similar to @defaultForInvalidEnumValue)?
     * with a special put() on B2JsonUnionTypeMap.Builder? putUnknownTypeHandler()?
     * other?

  * what restrictions are there on that type?
     * presumably there can be at most one type designated this way.
     * what fields is it allowed to have?
        * is it allowed to have any?
        * is it allowed to have any required fields?
        * other?

  * if there isn't one, do we just continue to throw a parse error?  (i'm assuming yes.)

bwbeach:
Unions now support default values that are used when the type is unknown.

PR #82, opened 2019-08-07
Address issue #46, and improve default value handling

bwbeach:
#46

You can now provide a default value to use in a union when an unknown type is found.

Also, default values are now deserialized each time they are used, which is safer if the values are mutable.

### DIFF
--- a/core/src/main/java/com/backblaze/b2/json/B2Json.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2Json.java
@@ -463,6 +463,17 @@ public class B2Json {
     }
 
     /**
+     * <p>Class annotation that applies to a class that is a @union.</p>
+     *
+     * <p>The value provided when de-serializing and the type is unknown.</p>
+     */
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.TYPE)
+    public @interface defaultForUnknownType {
+        String value();
+    }
+
+    /**
      * Field annotation that says a field is required to be present.
      */
     @Retention(RetentionPolicy.RUNTIME)
--- a/core/src/main/java/com/backblaze/b2/json/B2JsonHandlerMap.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2JsonHandlerMap.java
@@ -50,12 +50,6 @@ public class B2JsonHandlerMap {
     private final Stack<B2JsonInitializedTypeHandler> handlersToInitialize = new Stack<>();
 
     /**
-     * Handlers that need to have their default values checked.  This can only be done after
-     * all handlers are initialized, because the default values may require all of the handlers.
-     */
-    private final Stack<B2JsonInitializedTypeHandler> handlersToCheckDefaultValues = new Stack<>();
-
-    /**
      * Sets up a new map.
      */
     private B2JsonHandlerMap(Map<Class<?>, B2JsonTypeHandler<?>> initialMapOrNull) {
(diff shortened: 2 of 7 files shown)


core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -1542,6 +1561,80 @@ public void testSerializeUnregisteredUnionSubType() throws B2JsonException {
...
+        public int hashCode() {
+            return Objects.hash(n);
+        }
+    }
+
+    @Test
+    public void testUnionWithDefault() throws B2JsonException {
+        assertEquals(
+                new UnionWithDefaultClassA(5), // the default value
+                B2Json.get().fromJson("{\"type\": \"unknown\"}", UnionWithDefault.class)
```
certainmagic: add a test with an unknown value that has fields we don't expect?
for instance:

>{
>   "type" : "person",
>   "age" : 47,
>   "extra" : { "foo" : "bar" }
>}

this will make sure we can properly ignore unknown stuff inside the unknown type.

core/src/main/java/com/backblaze/b2/json/B2Json.java
```diff
@@ -462,6 +462,17 @@ public String listToJson(List<?> list, Class<?> valueClass, B2JsonOptions option
...
     }
 
+    /**
+     * <p>Class annotation that applies to a class that is a @union.</p>
+     *
+     * <p>The value provided when de-serializing and the type is unknown.</p>
+     */
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.TYPE)
+    public @interface defaultForUnknownType {
```
certainmagic: it's less a "default" and more a "replacementValue"....but this is completely consistent with B2Json.defaultForInvalidEnumValue, so....i guess it's fine.

core/src/main/java/com/backblaze/b2/json/B2JsonHandlerMap.java
```diff
@@ -89,9 +97,25 @@ private B2JsonHandlerMap(Map<Class<?>, B2JsonTypeHandler<?>> initialMapOrNull) {
...
         B2JsonTypeHandler<T> handler = getUninitializedHandler(clazz);
-        while (!handlersToInitialize.isEmpty()) {
-            handlersToInitialize.pop().initialize(this);
+
+        // If we created some handlers, then we need to initialize them and validate their
+        // default values.
+        //
+        // The reason for validating default values is so that we don't create a handler with
+        // bad defaults, and then not find out until much later when they happen to be used.
+        if (!handlersToInitialize.isEmpty()) {
```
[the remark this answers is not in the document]
bwbeach: Yep.

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -536,8 +570,18 @@ private T deserializeFromConstructorArgs(Object[] constructorArgs, int version)
...
+                    // We do a fresh deserialization of the default value each time, in case it's
+                    // a mutable type such as a List.
+                    try {
+                        constructorArgs[index] =
+                                fieldInfo.handler.deserialize(
+                                        new B2JsonReader(new StringReader(fieldInfo.defaultValueJsonOrNull)),
+                                        B2JsonOptions.DEFAULT
+                                );
+                    } catch (IOException e) {
+                        throw new B2JsonException(e.getMessage());
```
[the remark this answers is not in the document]
bwbeach:
```
                        // This should never happen.  We should have checked default values
                        // in B2JsonHandlerMap right after initializing the handlers and before
                        // anybody tried to use them.  OTOH, if deserializing a default value
                        // uses the default values in other types, we could hit this while
                        // checking default values in B2JsonHandlerMap.```

File: core/src/test/java/com/backblaze/b2/json/B2JsonTest.java (hunk 1 above)
Comment (bwbeach) [reply core/src/test/java/com/backblaze/b2/json/B2JsonTest.java]: Good catch!  It didn't work.  Added test, then added fix to not throw about an unknown field name in a union until the type is extracted, checked, and the default value is used if it's an unknown type.  Then it throws about an unknown field.

File: core/src/main/java/com/backblaze/b2/json/B2JsonUnionBaseHandler.java [hunk 5]
```diff
@@ -273,6 +273,12 @@ public T deserialize(B2JsonReader in, B2JsonOptions options) throws B2JsonExcept
 
         B2Preconditions.checkState(isInitialized());
 
+        // Place to hold the name of one of the unknow fields, if there are any.
```
certainmagic: unknown :)

PR #83, opened 2019-08-13
Fix bug with Sets contained in unions.

bwbeach:
When the `B2JsonUnionBaseHandler` gets the handler for a field, it
needs to use the same logic that `B2JsonObjectHandler` uses to create
the handler for a field.
    
This worked before the change to allow recursive unions because the fields
were pulled from the object handlers, which had already been initialized
when the union was set up.  The union can no longer assume that, so it
needs to get the un-initialized the field handlers itself.
    
Testing: new unit test

### DIFF
--- a/core/src/main/java/com/backblaze/b2/json/B2JsonUnionBaseHandler.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2JsonUnionBaseHandler.java
@@ -116,10 +116,10 @@ public class B2JsonUnionBaseHandler<T> extends B2JsonNonUrlTypeHandler<T> {
         fieldNameToHandler = new HashMap<>();
         final Map<String, String> fieldNameToSourceClassName = new HashMap<>();
         for (Class<?> subclass : typeNameToClass.values()) {
-            for (Field field : B2JsonObjectHandler.getObjectFieldsForJson(subclass)) {
+            for (Field field : B2JsonHandlerMap.getObjectFieldsForJson(subclass)) {
                 final String fieldName = field.getName();
                 final B2JsonTypeHandler handler =
-                        B2JsonObjectHandler.getUninitializedFieldHandler(
+                        B2JsonHandlerMap.getUninitializedFieldHandler(
                                 field.getGenericType(),
                                 b2JsonHandlerMap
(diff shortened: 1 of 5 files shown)


core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -176,6 +178,24 @@ public void testUnionWithTypeFieldNotLast() throws IOException, B2JsonException
         checkDeserializeSerialize(json, UnionAZ.class);
     }
 
+    /**
+     * Regression test to make sure that when handlers are created from
+     * B2JsonUnionHandler they work properly.
+     *
+     * This test makes a new B2Json, and deserializes the union type, so
+     * it's sure that the
```
certainmagic: cliff hanger?  :)

core/src/main/java/com/backblaze/b2/json/B2JsonUnionBaseHandler.java
```diff
@@ -118,7 +118,11 @@ protected void initializeImplementation(B2JsonHandlerMap b2JsonHandlerMap) throw
         for (Class<?> subclass : typeNameToClass.values()) {
             for (Field field : B2JsonObjectHandler.getObjectFieldsForJson(subclass)) {
                 final String fieldName = field.getName();
-                final B2JsonTypeHandler handler = b2JsonHandlerMap.getUninitializedHandler(field.getType());
+                final B2JsonTypeHandler handler =
```
certainmagic: i can see where this would be better.

...and it makes me wonder why we have two things so similar.
could the contents of the B2JsonObjectHandler version be
moved into the B2JsonHandlerMap version?

bwbeach: I'll move the methods to `B2JsonHandlerMap`, and add some comments.

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
bwbeach: Fixed.

PR #90, opened 2019-09-23
Adding an omitNull parameter to @B2Json.optional

malaysf:
Adding omitNull parameter to the @B2Json.optional annotation to not serialize fields that have a null value. This is false by default.

### DIFF
--- a/core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
+++ b/core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
@@ -2147,6 +2147,16 @@ public class B2JsonTest extends B2BaseTest {
                 B2Json.toJsonOrThrowRuntime(secureContainer, options));
     }
 
+    private static class OmitNullBadTestClass {
+        @B2Json.optional(omitNull = true)
+        private final int omitNullInt;
+
+        @B2Json.constructor(params = "omitNullInt")
+        public OmitNullBadTestClass(int omitNullInt) {
+            this.omitNullInt = omitNullInt;
+        }
+    }
+
     private static class OmitNullTestClass {
         @B2Json.optional(omitNull = true)
         private final String omitNullString;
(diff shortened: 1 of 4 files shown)


core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -2147,6 +2147,53 @@ public void testSensitiveWrittenWhenOptionNotSet() {
...
 
+    private static class OmitNullTestClass {
+        @B2Json.optional(omitNull = true)
+        private final String omitNullString;
+
+        @B2Json.optional
+        private final String regularString;
+
+        @B2Json.optional(omitNull = true)
+        private final int omitNullInt;
```
bwbeach: Would it be clearer to include a field of type `Integer`, which could be null?  (In addition to this field, of type `int`, which is also a good test.)

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -2147,6 +2147,53 @@ public void testSensitiveWrittenWhenOptionNotSet() {
...
+        @B2Json.constructor(params = "omitNullString, regularString, omitNullInt, regularInt")
+        public OmitNullTestClass(String omitNullString, String regularString, int omitNullInt, int regularInt) {
+            this.omitNullString = omitNullString;
+            this.regularString = regularString;
+            this.omitNullInt = omitNullInt;
+            this.regularInt = regularInt;
+        }
+    }
+
+    @Test public void testOmitNullWithNullInputs() {
```
bwbeach: A test that de-serializes `{}`, then reserializes, might be interesting.

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
certainmagic: should putting omitNull on a non-nullable field be an error at the time we parse the annotations?

malaysf: Implemented both suggestions

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
malaysf: implemented

PR #93, opened 2019-10-16
add options to bucket and keys

fmbz:


core/src/main/java/com/backblaze/b2/util/B2Collections.java
```diff
@@ -19,6 +19,10 @@
         return Collections.unmodifiableSet(set);
     }
 
+    public static <T> Set<T> emptySet() {
```
bwbeach: Would it be better to make a constant, and re-use the same instance?

core/src/main/java/com/backblaze/b2/client/structures/B2Bucket.java
```diff
@@ -97,6 +104,7 @@ public String toString() {
                 (bucketInfo == null ? 0 : bucketInfo.size()) + " infos," +
                 (corsRules == null ? 0 : corsRules.size()) + " corsRules," +
                 (lifecycleRules == null ? 0 : lifecycleRules.size()) + " lifecycleRules," +
+                (options == null ? 0 : options.size()) + " options," +
```
bwbeach: Wondering about listing the options here.  Would that be reasonable

core/src/main/java/com/backblaze/b2/client/structures/B2ApplicationKey.java
```diff
@@ -45,15 +49,17 @@
                     "capabilities, " +
                     "bucketId, " +
                     "namePrefix, " +
-                    "expirationTimestamp"
+                    "expirationTimestamp, " +
+                    "options"
     )
     public B2ApplicationKey(String accountId,
```
bwbeach: Incompatible change?  Does this need a major version number bump?

core/src/main/java/com/backblaze/b2/util/B2Collections.java
certainmagic: maybe [Collections.emptySet()](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#emptySet--) instead?

core/src/main/java/com/backblaze/b2/client/structures/B2ApplicationKey.java
```diff
@@ -36,6 +37,9 @@
     @B2Json.optional
     private final Long expirationTimestamp;
 
+    @B2Json.required
```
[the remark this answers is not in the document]
malaysf: Agree with ab, better to make this optional and not have to think about whether this is safe :)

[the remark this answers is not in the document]
fmbz: was an oversight - updated

core/src/main/java/com/backblaze/b2/util/B2Collections.java
fmbz: updated

core/src/main/java/com/backblaze/b2/client/structures/B2Bucket.java
fmbz: updated

core/src/main/java/com/backblaze/b2/client/structures/B2ApplicationKey.java
fmbz: updated

malaysf: the B2Json.constructor annotation can only be used once per class.

Looks like there isn't a unit test for this class. Fabian, can you please create one for it? Running that test should result in an error due to having two constructor annotations

certainmagic: i'm fine with bumping the version.  in fact, fabian & i just looked and we've bumped to v4 because of other little (breaking) changes and haven't released yet, so bumping is free.  :)

core/src/main/java/com/backblaze/b2/client/structures/B2Bucket.java
```diff
@@ -46,13 +49,38 @@ public B2Bucket(String accountId,
...
+                bucketType,
+                bucketInfo,
+                corsRules,
+                lifecycleRules,
+                null,
+                revision
+        );
+    }
+
+    @B2Json.constructor(params = "accountId,bucketId,bucketName,bucketType," +
```
certainmagic: as Malay noted above, we can only have on @B2Json.constructor annotation.

core/src/main/java/com/backblaze/b2/util/B2Collections.java
certainmagic: sorry.  i wasn't very clear.  maybe the callers could use Collections.emptySet() instead of us making our own wrapper for it.  i think i just made the B2Collections class to hold things I wanted but which weren't available in plain old java -- especially stuff I'd gotten used to having in guava, which we don't use in the sdk.

core/src/main/java/com/backblaze/b2/client/structures/B2ApplicationKey.java
certainmagic: re: unit test malay suggested -- one particularly good test is to make a structure, toJson it, fromJson it, and verify that the original and the copy match with equals.  you can also toJson again to be sure that the two json versions also match, but i'm not sure that's strictly necessary most of the time (at least not if you trust your equals method).

core/src/main/java/com/backblaze/b2/util/B2Collections.java
fmbz: updated

core/src/main/java/com/backblaze/b2/client/structures/B2ApplicationKey.java
fmbz: thanks for suggestion - updated

core/src/main/java/com/backblaze/b2/client/structures/B2Bucket.java
fmbz: updated

PR #94, opened 2019-11-05
Url cxn experiment

OldSkoolMark:
This PR contains both AB's B2WebApiUrlConnectionClientImpl and my B2AccountAuthorizerV5AuthImpl . It does not have the necessary changes to support Android pre API 26, which are significant, and required the elimination of some Java 8 language idioms and replacing the Java Date library with ThreeTenABP.

### DIFF
--- a/core/src/test/java/com/backblaze/b2/client/B2StorageClientFactoryPathBasedImplTest.java
+++ b/core/src/test/java/com/backblaze/b2/client/B2StorageClientFactoryPathBasedImplTest.java
@@ -8,6 +8,7 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.mockingDetails;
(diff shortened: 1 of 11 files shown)


core/src/test/java/com/backblaze/b2/client/B2StorageClientFactoryPathBasedImplTest.java
```diff
@@ -18,11 +19,10 @@
 
 
     @Test
-    public void testDefaultFactory_failsBecauseTestEnvironmentDoesntIncludeHttpClientJars() {
+    public void testDefaultFactory_failsSucceedsTestEnvironmentIncludesUrlBasedImplementation() {
```
certainmagic: would you please change "failsSucceeds" to just "succeeds", please?  (i'm pretty sure that's my mistake!)

samples/src/main/java/com/backblaze/b2/sample/B2Sample.java
```diff
@@ -420,8 +420,9 @@ private static void writeToFile(File file,
      */
 
 
+    @SuppressWarnings("SameParameterValue")
     private static File makeHugeFile(String fileNamePrefix) throws IOException {
-        final long fileSize = (long)Integer.MAX_VALUE + 1000*1000;
+        final long fileSize = 1000 * 1000 ;//(long)Integer.MAX_VALUE + 1000*1000;
```
certainmagic: is this a change i made?  do you know why?

PR #99, opened 2019-12-12
Add general generics support to B2Json.

generictjohnson:
This PR adds support for (de)serialization of class hierarchies that contain generics. As long as the root class provides all the context necessary, we should be able to resolve any nested types by propagating type parameters down.

This changes the mapping from class -> handler to type -> handler. The recursion is more streamlined (no more `static` `getUnitializedFieldHandler()` method). If handlers are tied to parameteric types, they are cached by class and type arguments to avoid collisions.

This PR is broken in many many small PRs, so it can be reviewed in small steps as well as by just comparing straight to master. Hopefully that is helpful. The unit tests pass at each intermediate commit.

### DIFF
--- a/core/src/main/java/com/backblaze/b2/json/B2JsonHandlerMap.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2JsonHandlerMap.java
@@ -232,10 +232,18 @@ public class B2JsonHandlerMap {
     /*package*/ synchronized <T> B2JsonTypeHandler<T> getUninitializedHandler(Type type) throws B2JsonException {
         // TODO validate the type - make sure it's resolved.
 
+        // class Item {
+        //     private int itemNumber;
+        //     private String itemName;
+        // }
         if (type instanceof Class) {
             final Class<T> clazz = (Class<T>) type;
             return getUninitializedHandlerForClass(clazz);
         }
+
+        // class Dataset {
+        //     private List<Double> measurements;
+        // }
         if (type instanceof ParameterizedType) {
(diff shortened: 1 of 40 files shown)


core/src/main/java/com/backblaze/b2/json/B2JsonHandlerMap.java
```diff
@@ -232,10 +232,18 @@ private B2JsonHandlerMap(Map<Class<?>, B2JsonTypeHandler<?>> initialMapOrNull) {
...
+        // }
         if (type instanceof Class) {
             final Class<T> clazz = (Class<T>) type;
             return getUninitializedHandlerForClass(clazz);
         }
+
+        // class Dataset {
+        //     private List<Double> measurements;
+        // }
         if (type instanceof ParameterizedType) {
```
bwbeach: The type here is from the field, right? It's the `List<Double>`?

Having the `class Dataset` in the comment above is a little confusing.

core/src/main/java/com/backblaze/b2/json/TypeResolver.java
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+
+        if (type instanceof TypeVariable) {
+            // If we're here, then type needs to be resolved.
+            // If there's no typeMap, then throw, because we cannot resolve anything.
+            if (typeMap == null) {
+                throw new RuntimeException("Cannot resolve type " + type + " - the typeMap is empty");
+            }
+            return typeMap.get(type.getTypeName());
```
bwbeach: throw if result is null?

PR #101, opened 2020-01-13
Adding support to override additional download headers.

malaysf:
These are specified in the requests to b2_get_download_authorization, b2_download_file_by_id, and b2_download_file_by_name

### DIFF
--- a/core/src/main/java/com/backblaze/b2/client/B2StorageClientWebifierImpl.java
+++ b/core/src/main/java/com/backblaze/b2/client/B2StorageClientWebifierImpl.java
@@ -412,7 +412,7 @@ public class B2StorageClientWebifierImpl implements B2StorageClientWebifier {
                              B2DownloadByIdRequest request,
                              B2ContentSink handler) throws B2Exception {
         downloadGuts(accountAuth,
-                makeDownloadByIdUrl(accountAuth, request.getFileId(), request),
+                makeDownloadByIdUrl(accountAuth, request),
                 request.getRange(),
                 handler);
     }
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/backblaze/b2/client/B2StorageClientWebifierImpl.java
```diff
@@ -599,27 +614,58 @@ private String makeGetFileInfoByNameUrl(B2AccountAuthorization accountAuth,
...
+        uriBuilder
+                .append("file/")
+                .append(bucketName)
+                .append("/")
+                .append(percentEncode(fileName));
+
+        if (request != null) {
+            int queryparameterCount = 0;
+
+            queryparameterCount = maybeAddQueryParamToUrl(uriBuilder, queryparameterCount, "b2ContentDisposition", request.getB2ContentDisposition());
```
certainmagic: someday we might want to have an interface for request objects that has these header accessor methods so we can share the code that adds them to the urls...?

core/src/main/java/com/backblaze/b2/client/structures/B2DownloadByIdRequest.java
```diff
@@ -67,13 +112,45 @@ public Builder setRange(B2ByteRange range) {
...
             this.b2ContentDisposition = b2ContentDisposition;
             return this;
         }
+
+        public Builder setB2ContentLanguage(String b2ContentLanguage) {
+            this.b2ContentLanguage = b2ContentLanguage;
+            return this;
+        }
+
+        public Builder setB2Expires(String b2Expires) {
```
certainmagic: it might be nice to provide a convenience version of this that takes a LocalDateTime object, assumed to be in UTC.

PR #109, opened 2020-04-03
B2Json: Allow generic types for top-level objects from select entry points

generictjohnson:
This allows the user to supply type information to aid in `B2JsonTypeHandler` resolution for the top-level object. If the type information is not supplied, `B2Json` falls back to the existing behavior of resolving by class.

The first commit is a small refactor.
Second is the feature.
Third is a test.

### DIFF
--- a/core/src/main/java/com/backblaze/b2/json/B2Json.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2Json.java
@@ -169,15 +169,8 @@ public class B2Json {
     }
 
     public String toJson(Object obj, B2JsonOptions options) throws B2JsonException {
-        if (obj == null) {
-            throw new B2JsonException("top level object must not be null");
-        }
-        Class<?> clazz = obj.getClass();
-        final B2JsonTypeHandler handler = handlerMap.getHandler(clazz);
         try (final ByteArrayOutputStream out = new ByteArrayOutputStream()) {
-            B2JsonWriter jsonWriter = new B2JsonWriter(out, options);
-            //noinspection unchecked
-            handler.serialize(obj, options, jsonWriter);
+            toJson(obj, options, out);
             return out.toString(B2StringUtil.UTF8);
         } catch (IOException e) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/backblaze/b2/json/B2Json.java
```diff
@@ -139,23 +140,35 @@ public static B2Json get() {
         }
     }
 
+    public void toJson(Object obj, OutputStream out) throws IOException, B2JsonException {
+        toJson(obj, B2JsonOptions.DEFAULT, out);
+    }
+
+    public void toJson(Object obj, B2JsonOptions options, OutputStream out) throws IOException, B2JsonException {
+        toJson(obj, options, out, null);
+    }
```
bwbeach, line 149: How about javadocs for these two methods?  They are public methods that users of the library will be calling.

core/src/main/java/com/backblaze/b2/json/B2Json.java
```diff
@@ -139,23 +140,35 @@ public static B2Json get() {
...
+
     /**
      * Turn an object into JSON, writing the results to given
      * output stream.
      *
+     * objTypeOrNull can be set to null if obj is not a parameterized class. However,
+     * if obj contains type parameters (like if obj is a {@literal List<String>}, then
+     * you will need to pass in its type information via objTypeOrNull. This will instruct
+     * B2Json to derive the B2JsonTypeHandler from the type information instead of the
+     * object's class.
```
bwbeach, line 159: Maybe give an example of how to get the type handler if I have a `List<String>`?

core/src/main/java/com/backblaze/b2/json/B2Json.java
generictjohnson: Done

core/src/main/java/com/backblaze/b2/json/B2Json.java
generictjohnson: Done

PR #112, opened 2020-04-17
validate chars in file info header name

fmbz:
This PR validates against all supported chars in RFC 7230 https://tools.ietf.org/html/rfc7230#section-3.2

Testing: All tests


certainmagic:
thanks, @fmbz.
could you please put the reference to the RFC & its link back in? and then i think we're good to go.  :)

PR #123, opened 2020-08-21
Fix regular expression that has an unescaped

dwisiswant0:
This regular expression has an unescaped `.`, so it might match more hosts than expected. For example, this lines below will show LGTM.

```python
######################################################################
#
# File: check_licenses
#
# Copyright 2017, Backblaze Incx All Rights Reserved.
#
# License https://wwwxbackblazexcom/using_b2_codexhtml
#
######################################################################
```

```bash
> ./check_code check_code
check_code: looks good.
```

### DIFF
--- a/check_code
+++ b/check_code
@@ -37,8 +37,8 @@ IN_LIB_RE = re.compile("/lib/[^/]+$")
 #
 ########################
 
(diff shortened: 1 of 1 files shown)


certainmagic (approved):
Thanks, Dwi!  :)

Welcome to the SDK.  :)

PR #130, opened 2020-11-13
support more than 64 fields per object.

a3eu:
replaced bits in a long with a bit set.

### DIFF
--- a/core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
+++ b/core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
@@ -430,7 +430,6 @@ public class B2JsonObjectHandler<T> extends B2JsonTypeHandlerWithDefaults<T> {
         final Object [] constructorArgs = new Object [fields.length];
 
         // Read the values that are present in the map.
-        final BitSet foundFieldBits = new BitSet();
         if (fieldNameToValue == null) {
             throw new B2JsonException("B2JsonObjectHandler.deserializeFromFieldNameToValueMap called with null fieldNameToValue");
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -486,7 +486,7 @@ public T deserializeFromUrlParameterMap(Map<String, String> parameterMap, B2Json
                     throw new B2JsonException("required field " + fieldInfo.getName() + " cannot be null");
                 }
                 constructorArgs[fieldInfo.constructorArgIndex] = value;
-                foundFieldBits |= fieldInfo.bit;
+                foundFieldBits.set(fieldInfo.constructorArgIndex);
```
ericjding, line 489: Alex, ab and I had discussed that it seems like this code sets the bits but doesn't actually use them (unlike in deserialize(). Could you look through and see if there's some missing _usage_ of the bits and add as needed while you're in here?

ericjding (approved):
Looks good, please see my comment for possible additional change.

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -399,7 +399,7 @@ public T deserialize(B2JsonReader in, B2JsonOptions options) throws B2JsonExcept
                     in.skipValue();
                 }
                 else {
-                    if ((foundFieldBits & fieldInfo.bit) != 0) {
+                    if (foundFieldBits.get(fieldInfo.constructorArgIndex)) {
```
certainmagic, line 402: is foundFieldBits used in the other methods?  should it be?

varikin (approved):

PR #131, opened 2020-11-13
Use Python 3.6 for scripts

keithele:
Update to use Python 3.6 instead of Python 2.7 for Travis-CI.


keithele:
Moving to travis-ci.com and other updates have fixed this issue.

PR #133, opened 2020-11-19
pip3

tlucas30:
b2 was still using python 2.7 pip3 should fix that fingers crossed


malaysf (approved):
Looks fine and the build succeeded

PR #142, opened 2021-01-13
disable content compression so as not to automatically decompress compressed content

mxue-BB:
- Disable content compression via client RequestConfig so getContent will not automatically decompress compressed content from server side
- add some unit test for B2WebApiHttpClient

Jira: https://backblaze.atlassian.net/browse/DEV-6545


httpclient/src/main/java/com/backblaze/b2/client/webApiHttpClient/B2WebApiHttpClientImpl.java
```diff
@@ -102,6 +103,20 @@ public void getContent(String url,
             get.setHeaders(makeHeaders(headersOrNull));
         }
 
+        /* At default HTTP client libraries will automatically decompress compressed content
```
malaysf, line 106: `at default` -> `by default,`

mxue-BB: corrected it. thanks.

malaysf (approved):

PR #143, opened 2021-01-19
create B2Json version of ByteArrayOutputStream and throws IOException…

mxue-BB:
- create B2Json version of ByteArrayOutputStream for B2Json API
- throws IOException("Requested array size exceeds maximum limit") OutputStream requests array size over the limit (Integer.MAX_VALUE - 8)

Notes: 
A subclass of B2JsonByteArrayOutputStream was created for testing with a smaller MAX_ARRAY_SIZE for throwing IOException("Requested array size exceeds maximum limit"). The original MAX_ARRAY_SIZE(Integer.MAX_VALUE - 8) was so large that we would throw an OutOfMemoryError first before even getting a chance to throw IOException with most typical JVM heap settings. (I managed to throw IOException with a larger heap size (>=3 GB) for unit testing B2JsonByteArrayOutputStream with it original MAX_ARRAY_SIZE)
The above fact means using B2JsonByteArrayOutputStream to catch and throw the intended IOException is of practical value probably only for JVMs with very large heap size.


core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -2792,6 +2793,104 @@ public void testClassWithLotsOfFields() throws IOException, B2JsonException {
...
+
+       @Override
+       public int hashCode() {
+           return name.hashCode();
+       }
+    }
+
+    @Test
+    public void testObjectWithNameToJson() throws B2JsonException, IOException {
+        // an object not big enough to cause IOException("Requested array size exceeds maximum limit")
```
certainmagic, line 2826: how does this differ from the other tests in this file where we're not overflowing the limit?

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -2792,6 +2793,104 @@ public void testClassWithLotsOfFields() throws IOException, B2JsonException {
...
+        }
+        // the total size of object in bytes is 494; doubling will not cause overflow over 1000
+        final ObjectWithSomeName expectedObjectWithSomeSmallName = new ObjectWithSomeName(smallName.toString());
+        final B2JsonByteArrayOutputStreamForTest b2JsonByteArrayOutputStreamForSmall = new B2JsonByteArrayOutputStreamForTest();
+        B2Json.get().toJson(expectedObjectWithSomeSmallName, b2JsonByteArrayOutputStreamForSmall);
+
+        // rebuild the object based on the Json string
+        final String actualJson = b2JsonByteArrayOutputStreamForSmall.toString();
+        final ObjectWithSomeName actualObjectWithSomeName = B2Json.fromJsonOrThrowRuntime(actualJson, ObjectWithSomeName.class);
+        assertEquals(expectedObjectWithSomeSmallName, actualObjectWithSomeName);
```
certainmagic, line 2841: btw, if we get rid of this, we can probably also get rid of equals() and hashCode() on the test class.

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -2792,6 +2793,104 @@ public void testClassWithLotsOfFields() throws IOException, B2JsonException {
...
+     * close to half of MAX_ARRAY_SIZE.
+     *
+     * Instead, create a subclass of B2JsonByteArrayOutputStream with lowered
+     * MAX_ARRAY_SIZE (1000) for testing purpose: IOException will then be thrown
+     */
+    private static class B2JsonByteArrayOutputStreamForTest extends B2JsonByteArrayOutputStream {
+        private static final int MAX_ARRAY_SIZE = 1000;
+
+        @Override
+        protected void grow(int minCapacity) throws IOException {
```
certainmagic: i like that you've got a way of testing the boundary condition in a unit test.

could we do it without having to make a copy of grow() which may get out-of-sync with the real implementation?

maybe we should make "maxCapacity" be a member and have it default to Integer.MAX_VALUE.  And this subclass would just pass a non-default maxCapacity?  yes, it would take a little bit of extra space, but it's not like we have millions of these at a time and it _does_ make it more testable.

whatcha think? got a better proposal?  :)

core/src/main/java/com/backblaze/b2/json/B2JsonByteArrayOutputStream.java
```diff
@@ -0,0 +1,219 @@
+/*
+ * Copyright 2021, Backblaze Inc. All Rights Reserved.
```
certainmagic, line 2: This code looks remarkably similar to the code in ByteArrayOutputStream.java that i see in the IDE.  Similar enough that I don't think we actually wrote it and we need to figure something out.  Where did the code come from and what did we change?  (And why do we think that's reasonable?)

core/src/test/java/com/backblaze/b2/json/B2JsonByteArrayOutputStreamTest.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Unit test for B2JsonByteArrayOutputStream
+ */
+public class B2JsonByteArrayOutputStreamTest {
+
+    @Rule
+    public ExpectedException thrown  = ExpectedException.none();
+
+    @Test
+    public void testRequestedArraySizeOverTheMaxLimit() throws IOException {
```
certainmagic, line 54: after we figure out what implementation will use for B2JsonByteArrayOutputStream, let's think about what we have to test here.

core/src/main/java/com/backblaze/b2/json/B2Json.java
```diff
@@ -193,11 +201,11 @@ public String toJson(Object obj) throws B2JsonException {
     }
 
     public String toJson(Object obj, B2JsonOptions options) throws B2JsonException {
-        try (final ByteArrayOutputStream out = new ByteArrayOutputStream()) {
+        try (final B2JsonByteArrayOutputStream out = new B2JsonByteArrayOutputStream()) {
             toJson(obj, options, out);
             return out.toString(B2StringUtil.UTF8);
         } catch (IOException e) {
-            throw new RuntimeException("IO exception writing to string");
+            throw new RuntimeException("IO exception writing to string: " + e.getMessage());
```
certainmagic, line 208: i _think_ that B2Json is careful to not echo a user's content back in an exception....and that _might_ be why we weren't including the IOException's message in the RuntimeException's message.  can you please take a look to see if we're including the underlying message anywhere else that _might_ contain user-provided content?

core/src/main/java/com/backblaze/b2/json/B2Json.java
```diff
@@ -142,16 +144,20 @@ public static B2Json get() {
 
     /**
      * Turn an object into JSON, writing the results to given output stream.
+     *
+     * (package access for testing)
      */
-    public void toJson(Object obj, OutputStream out) throws IOException, B2JsonException {
+    void toJson(Object obj, OutputStream out) throws IOException, B2JsonException {
```
certainmagic, line 150: why would we _remove_ "public" for testing purposes?  i can see removing "private", but I think this change is altering the SDK's API.

certainmagic: same question for the other instances below.

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
```diff
@@ -2792,6 +2793,104 @@ public void testClassWithLotsOfFields() throws IOException, B2JsonException {
...
+       @Override
+       public int hashCode() {
+           return name.hashCode();
+       }
+    }
+
+    @Test
+    public void testObjectWithNameToJson() throws B2JsonException, IOException {
+        // an object not big enough to cause IOException("Requested array size exceeds maximum limit")
+        final StringBuilder smallName = new StringBuilder();
```
certainmagic, line 2827: maybe make a helper method makeStringWithLength() to be clearer?

certainmagic (changes requested):

core/src/main/java/com/backblaze/b2/json/B2JsonByteArrayOutputStream.java
```diff
@@ -0,0 +1,219 @@
...
+ * THREAD-SAFE
+ */
+public class B2JsonByteArrayOutputStream extends OutputStream {
+
+    /**
+     * The maximum size of array to allocate. Reserve some header words
+     * in an array. Attempts to allocate larger arrays will result in
+     * IOException: Requested array size exceeds maximum limit.
+     */
+    private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
```
malaysf, line 26: How did you arrive at `Integer.MAX_VALUE - 8` being the proper maximum that we should support?

Should this be configurable? Integer.MAX_VALUE is pretty large, it may be too large for many applications that don't configure huge heaps for Java

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
mxue-BB: thinking about it more: we can get rid of this part.

core/src/test/java/com/backblaze/b2/json/B2JsonTest.java
mxue-BB: yes

core/src/main/java/com/backblaze/b2/json/B2JsonByteArrayOutputStream.java
mxue-BB: this value is used in the original version of ByteArrayOutputStream. There is not much detail about it in that version. I'd agree this definitely looks to be too large since we typically get into OOM error without triggering IOException (I did manage to get IOException with 2 GB heap). But I don't know what would be a reasonable value for this yet.

core/src/main/java/com/backblaze/b2/json/B2JsonByteArrayOutputStream.java
mxue-BB: Ops, not intentionally, I'd have to take that default copyright off. It is virtually a copy of standard version of ByteArrayOutputStream (minor changes of a couple of methods: grow and hugeCapacity). The thinking was: we were using the standard version of ByteArrayOutputStream and we could have an almost identical version of it with changed way of detecting the array allocation overflow and throwing IOException (as opposed OOM error). Somehow along the way the max array size (Integer.MAX_VALUE - 8) seems to be too big. Need to find a better max size limit.

core/src/main/java/com/backblaze/b2/json/B2Json.java
mxue-BB: there were 2 places that do get messages from the caught IO exceptions: toJsonUtf8Bytes, toJsonUtf8BytesWithNewline. Any particular reason for not doing that?

core/src/main/java/com/backblaze/b2/json/B2Json.java
mxue-BB: So far the methods with the public removed are used for testing only. I have not found that they are used as API. Another reason for making them non-public is to prevent the caller using the OOM-prone OutputStream in making these API calls.

core/src/test/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStreamTest.java
```diff
@@ -0,0 +1,67 @@
...
+import static junit.framework.Assert.assertNull;
+import static junit.framework.TestCase.assertEquals;
+
+/**
+ * Unit test for B2JsonByteArrayOutputStream
+ */
+public class B2JsonBoundedByteArrayOutputStreamTest {
+
+    /**
+     * Cannot directly test the case with MAX_ARRAY_SIZE being (Integer.MAX_VALUE - 8)
```
bwbeach, line 22: It would be cleaner to pass the maximum into the constructor.  I think having the caller explicitly specify the max is good, then the callers know what's going on.

core/src/main/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStream.java
```diff
@@ -0,0 +1,137 @@
...
+     * @param bytes input array of bytes
+     * @param offset offset for the input array
+     * @param length number of bytes to write
+     * @throws IOException if new expanded capacity is over
+     *                     the MAX_CAPACITY
+     */
+    @Override
+    public synchronized void write(byte[] bytes, int offset, int length) throws IOException {
+        // expand capacity if necessary
+        if (size + length > output.length) {
```
bwbeach, line 72: maybe factor out `size + length` -- it's computed twice, and giving it a name would be nice

core/src/main/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStream.java
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * expands the current capacity to meet the least required capacity
+     *
+     * @param  leastCapacityRequired the least required capacity to hold all content
+     * @return the new capacity
+     * @throws IOException if new expanded capacity is over the MAX_CAPACITY
+     */
+    private int expandCapacity(int leastCapacityRequired) throws IOException {
+        // double current capacity
+        int newCapacity = output.length * 2;
```
bwbeach, line 124: might be clearer to use `Math.max` rather than the `if` statement.

core/src/test/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStreamTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.io.IOException;
+
+import static junit.framework.Assert.assertNull;
+import static junit.framework.TestCase.assertEquals;
+
+/**
+ * Unit test for B2JsonByteArrayOutputStream
+ */
+public class B2JsonBoundedByteArrayOutputStreamTest {
```
bwbeach, line 19: Does this test have full coverage of the code?

In addition to full coverage, it would be good to test:

- empty results
- writing empty arrays
- boundary conditions at expansion size
- multiple expansions
- a result that is exactly the maximum size, written as one block
- a result that is exactly the maximum size, written one byte at a time

core/src/main/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStream.java
```diff
@@ -0,0 +1,137 @@
...
+    // byte array to hold output content
+    private byte[] output;
+
+    // the current number of bytes written so far
+    private int size;
+
+    // maximum capacity the output array is allowed to grow to
+    private static final int MAX_CAPACITY = Integer.MAX_VALUE - 8;
+
+    public B2JsonBoundedByteArrayOutputStream() {
```
bwbeach, line 33: As mentioned below, I think the constructor should take the maximum buffer size, rather than using a constant.

That will make it easier to read calling code and know what it's doing.

core/src/test/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStreamTest.java
mxue-BB: Added full coverage on public methods, and additional tests you mentioned (pretty much I think).

core/src/main/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStream.java
```diff
@@ -13,45 +15,51 @@
...
  * - stores stream content in a byte array
  * - expands capacity when needed (typically doubles)
- * - has a max capacity threshold
- * - throws IOException if the threshold is reached
+ * - has an upper bound limit on output array size (max capacity)
+ * - throws IOException if the max capacity threshold is crossed
  *
  * THREAD-SAFE
  */
 public class B2JsonBoundedByteArrayOutputStream extends OutputStream {
```
mxue-BB, line 30: We'd expose B2JsonBoundedByteArrayOutputStream as public so bzmono can replace standard ByteArrayOutputStream with this type of OutputStream as parameters to some B2Json APIs (to catch the intended IOException). For instance, B2Json.toJson(Object obj, B2JsonOptions options, OutputStream out).

core/src/test/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStreamTest.java
bwbeach: Is that full coverage of the entire class?  All the lines in the source file?

core/src/test/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStreamTest.java
```diff
@@ -0,0 +1,171 @@
...
+        assertArrayEquals(new byte[]{'A'}, b2JsonByteArrayOutputStream.toByteArray());
+        assertEquals(1, b2JsonByteArrayOutputStream.getSize());
+
+        b2JsonByteArrayOutputStream.write('B');
+        assertArrayEquals(new byte[] {'A', 'B'}, b2JsonByteArrayOutputStream.toByteArray());
+        assertEquals(2, b2JsonByteArrayOutputStream.getSize());
+
+        // create the actual bytes array to compare to the expected bytes later
+        final byte[] actualAllBytes = new byte[2048];
+        actualAllBytes[0] = 'A';
```
bwbeach, line 49: Why tread the first two bytes differently?  Maybe just start the loop below at 0?

core/src/test/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStreamTest.java
```diff
@@ -0,0 +1,171 @@
...
+        try {
+            b2JsonBoundedByteArrayOutputStream = new B2JsonBoundedByteArrayOutputStream(Integer.MAX_VALUE - 7);
+        } catch (IllegalArgumentException illegalArgumentException) {
+            assertEquals("maxCapacity must not be bigger than 2147483639", illegalArgumentException.getMessage());
+        }
+        assertNull(b2JsonBoundedByteArrayOutputStream);
+    }
+
+    /* append byte array 2 to the byte array 1 */
+    private byte[] appendOneByteArrayToTheOtherByteArray(byte[] byteArray1, byte[] byteArray2) {
```
bwbeach, line 158: `BzArrayUtil.appendByteArrays` does this.

core/src/test/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStreamTest.java
```diff
@@ -0,0 +1,171 @@
...
+    public void testInitialState() {
+        final B2JsonBoundedByteArrayOutputStream b2JsonByteArrayOutputStream = new B2JsonBoundedByteArrayOutputStream(128);
+        assertArrayEquals(new byte[0], b2JsonByteArrayOutputStream.toByteArray());
+        assertEquals(128, b2JsonByteArrayOutputStream.getMaxCapacity());
+        assertEquals(0, b2JsonByteArrayOutputStream.getSize());
+    }
+
+    @Test
+    public void testWriteOneByte() throws IOException {
+        final B2JsonBoundedByteArrayOutputStream b2JsonByteArrayOutputStream = new B2JsonBoundedByteArrayOutputStream(2048);
```
bwbeach, line 37: I think this would fail if the max were not a power of two.  that would be worth testing, too.

bwbeach (commented):
There's still a bug with using max sizes that aren't powers of 2.

And... have you resolved ab's issue about exposing messages in IOExceptions?

core/src/main/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStream.java
```diff
@@ -0,0 +1,180 @@
...
+     *
+     * @param bytes input array of bytes
+     * @param offset offset for the input array
+     * @param length number of bytes to write
+     * @throws IOException if expanding capacity would cross
+     *                     the maxCapacity threshold
+     */
+    @Override
+    public synchronized void write(byte[] bytes, int offset, int length) throws IOException {
+        // note: newSize could be negative due to overflow, check it in checkCapacity()
```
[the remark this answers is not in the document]
mxue-BB: 1) I had looked it System.arrayCopy and it throws exceptions on negative length/offset, etc. So I decided to skip it. But it is probably good idea to catch it here early and explicitly. 
2) I thought the current way is simpler but probably not very clearer, but I'll reconsider it.

core/src/main/java/com/backblaze/b2/json/B2Json.java
certainmagic: thanks for looking.  good question.  i think i don't remember what these messages tend to look like.  we probably thought that they didn't include incoming content and were instead messages like "disk full" or "permission denied".  messages like those seem safe.  the exception you added should be safe.  i looked though and java.io.UnsupportedEncodingException is an IOException and i don't know if it echos back content.

i wish that there wasn't such a thing as "malicious user content".   perhaps it's ok to include this and say it's the responsibility of higher-level software to not leak content (malicious or otherwise).

certainmagic: just checked with brianb.  he agrees it's ok to let these through.  this isn't really a problem we can completely solve for software above us.

core/src/main/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStream.java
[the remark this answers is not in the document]
mxue-BB: @bwbeach Yes, I've added the preconditions for index bound check and updated the check for overflow.

core/src/test/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStreamTest.java
```diff
@@ -0,0 +1,197 @@
...
+        for (int i = 0; i < 2048; i++) {
+            b2JsonByteArrayOutputStream.write(i);
+            actual2048Bytes[i] = (byte)i;
+        }
+
+        // OutputStream's array size and content should match actualAllBytes array
+        assertArrayEquals(actual2048Bytes, b2JsonByteArrayOutputStream.toByteArray());
+        assertEquals(2048, b2JsonByteArrayOutputStream.getSize());
+
+        // if the max capacity is not a power of 2 we can only write up to
```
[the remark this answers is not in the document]
mxue-BB: @bwbeach I've updated the code to address this bug and also update the unit test.

core/src/main/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStream.java
```diff
@@ -0,0 +1,185 @@
...
+     * @return output output array size
+     */
+    public int getSize() {
+        return this.size;
+    }
+
+    /**
+     * checks and expands capacity if needed
+     *
+     * @param length the new input byte length
```
bwbeach, line 141: This could be clearer.  I think this parameter is the additional capacity needed, in addition to the current size (not the current capacity).

bwbeach (approved):

core/src/main/java/com/backblaze/b2/json/B2JsonBoundedByteArrayOutputStream.java
mxue-BB: Right, updated.

certainmagic:
i just skimmed it.  if brianb is fine with it, i'm ok with it.

PR #153, opened 2021-06-09
Add s3ApiUrl to B2AccountAuthorization structure

johnlcox:
The `b2_authorize_account` API call includes the `s3ApiUrl` field in the response. This adds it to the JSON structure class. I tested this by using the SDK to make a `b2_authorize_account` call and confirmed it included the correct value for `s3ApiUrl`.


ericjding (approved):
LGTM

PR #156, opened 2021-09-17
Dev 10032

rhryckewicz:
changing from travis to github actions


.github/workflows/.ci_cd.yml
```diff
@@ -0,0 +1,96 @@
...
+  pull_request:
+    branches:
+    - master
+
+env:
+  PYTHON_DEFAULT_VERSION: 3.6
+  OUTPUT_DIR: $GITHUB_WORKSPACE/build/outputs
+  OUTPUT_ZIP: b2-sdk-build-${GITHUB_RUN_NUMBER}.zip
+  # This token was generated under rhryckewicz account
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
varikin, line 17: Now that you've tested out that this token is needed, lets get one from a service account. Can also be done in a follow up PR, well removing the comment would be a follow up PR :)

varikin (approved):

ericjding (approved):
Are we just testing a new flow/infrastructure? Then... approved!

PR #168, opened 2022-05-06
Add configuration for deploying to Maven Central through Gradle

DenWav:
Note for initial testing this only calls `closeSonatypeStagingRepository` so it will not fully deploy to Maven Central. Once we have confirmed the configuration here is correct, then we can replace that with `closeAndReleaseSonatypeStagingRepository` to fully automate the deployment.


malaysf (approved):

fmbz (approved):
LGTM

PR #170, opened 2022-05-11
Implement `createBundle` task in Gradle to create GitHub release bundle

DenWav:
The bundle jars which are uploaded to GitHub include all of the
artifacts including the signing files with the artifacts.


fmbz (approved):
LGTM

PR #171, opened 2022-05-16
Update to commons-codec 1.15

ooola:


httpclient/build.gradle.kts
```diff
@@ -17,5 +17,11 @@ dependencies {
     api(projects.b2SdkCore)
 
     api("org.apache.httpcomponents:httpclient:4.5.13")
+    constraints {
```
nschile, line 20: Let's also update the CHANGELOG.md with this change.

PR #177, opened 2022-08-31
Inference for `B2Json.constructor#params`

clamothebb:
Adds support for Java 8's `-parameters` option, so constructor parameters do not need to be reiterated in `B2Json.constructor#params`. This enables DRYer B2Json usage and allows developers to generate complete constructors for POJOs using IDE tools.

Previous syntax (still supported):
```java
        @B2Json.constructor(params = "omitNullString, regularString, omitNullInteger, regularInteger")
        public OmitNullTestClass(String omitNullString, String regularString, Integer omitNullInteger, Integer regularInteger) {
            this.omitNullString = omitNullString;
            this.regularString = regularString;
            this.omitNullInteger = omitNullInteger;
            this.regularInteger = regularInteger;
        }
```

New syntax supported:
```java
        @B2Json.constructor
        public OmitNullTestClass(String omitNullString, String regularString, Integer omitNullInteger, Integer regularInteger) {
            this.omitNullString = omitNullString;
            this.regularString = regularString;
            this.omitNullInteger = omitNullInteger;
            this.regularInteger = regularInteger;
        }
```

This doesn't introduce any dependencies. It continues to support classes not compiled with javac's `-parameters` argument. This doesn't have any effect on existing classes, since inference is only used when `params` are not provided.

Locally, I modified and ran the entire `B2JsonTest` suite with inferred parameters, with the exception of `testParamListedTwice` since it's impossible to list a param twice in an actual constructor. All tests passed. This PR includes focused tests for this new functionality.


core/src/main/java/com/backblaze/b2/json/B2Json.java
```diff
@@ -577,8 +577,9 @@ public <T> T fromUrlParameterMap(Map<String, String> parameterMap, Class<T> claz
      * should use.  This constructor must take ALL of the serializable
      * fields as parameters.
      *
-     * You must provide an "params" parameter that lists the order of
-     * the parameters to the constructor.
+     * You must provide a "params" parameter that lists the order of
```
eding-backblaze, line 580:
```suggestion
     * You must either provide a "params" parameter that lists the order of
```

core/src/main/java/com/backblaze/b2/json/B2Json.java
```diff
@@ -577,8 +577,9 @@ public <T> T fromUrlParameterMap(Map<String, String> parameterMap, Class<T> claz
...
+     * the parameters to the constructor or compile classes with the
```
eding-backblaze, line 581:
```suggestion
     * the parameters to the constructor or else compile classes with the
```

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -162,21 +159,38 @@ protected void initializeImplementation(B2JsonHandlerMap handlerMap) throws B2Js
...
-            final int constructorParamCount = fields.length + numberOfVersionParams;
-            if (paramNames.length != constructorParamCount) {
+            // Verify that, if present, the number of params specified in the annotation is correct
+            final int expectedParamCount = fields.length + numberOfVersionParams;
+            if (annotationParamNames != null && annotationParamNames.length != expectedParamCount) {
+                throw new IllegalArgumentException(clazz.getName() + " constructor's @B2Json.constructor annotation does not have the right number of params.");
+            }
+
+            // Verify that the number of actual constructor params is correct
+            Parameter[] constructorParams = this.constructor.getParameters();
```
eding-backblaze, line 176: Does this work even if javac is not called with the `-parameters` argument?

core/src/test/java/com/backblaze/b2/json/B2JsonInferredParametersTest.java
```diff
@@ -0,0 +1,2858 @@
+/*
+ * Copyright 2018, Backblaze Inc. All Rights Reserved.
```
eding-backblaze, line 2: This should probably be 2022, right?

core/src/test/java/com/backblaze/b2/json/B2JsonInferredParametersTest.java
```diff
@@ -0,0 +1,2858 @@
...
+            }
+            Container other = (Container) o;
+            return a == other.a && (b == null ? other.b == null : b.equals(other.b));
+        }
+    }
+
+    private static final B2Json b2Json = B2Json.get();
+
+    @Test
+    public void testBoolean() throws IOException, B2JsonException {
```
eding-backblaze, line 75: Do these tests (other than those that use your `Container` nested class and the others that don't use `@param`) somehow test something new? I'm not clear how they have anything to do with your newly introduced functionality.

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
clamothebb: Yes. If that option is removed from  `b2sdk.gradle.kts`, all tests pass except those in `B2JsonInferredParametersTest`, which fail with "[class] constructor is missing 'params' for its @B2Json.constructor annotation. Either specify this or add -parameters to javac args.

When `-parameters` is not enabled, the only differences that affect our usage are: 
* `Parameter#isNamePresent()` returns false
* `Parameter#getName()` returns `arg0`, `arg1`, etc.

core/src/test/java/com/backblaze/b2/json/B2JsonInferredParametersTest.java
eding-backblaze: Oh I see. So they test that your new functionality didn't indirectly break something else?

clamothebb: Yes. They test that all the existing (tested) features play nicely with this new functionality. Think of it as way to test for harmful 'drug interactions' with no medical knowledge. I'm new to this library, so don't have an intuition for which features are more likely to interact poorly with this. Instead, just test them all.

Now that we're shown there to be no poor interactions, having just a few focused unit tests will keep this more maintainable.

clamothebb: Done

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
```diff
@@ -162,21 +159,38 @@ protected void initializeImplementation(B2JsonHandlerMap handlerMap) throws B2Js
...
+                annotationParamNames = null;
             }
 
-            final int constructorParamCount = fields.length + numberOfVersionParams;
-            if (paramNames.length != constructorParamCount) {
-                throw new IllegalArgumentException(clazz.getName() + " constructor does not have the right number of parameters");
+            // Verify that, if present, the number of params specified in the annotation is correct
+            final int expectedParamCount = fields.length + numberOfVersionParams;
+            if (annotationParamNames != null && annotationParamNames.length != expectedParamCount) {
+                throw new B2JsonException(clazz.getName() + " constructor's @B2Json.constructor annotation does not have the right number of params.");
```
eding-backblaze, line 172: I think I like the change to `B2JsonException` but maybe @brianb-backblaze had a reason to use `IllegalArgumentException` originally...?

eding-backblaze (approved):
LGTM, but maybe wait for @brianb-backblaze to take another look too?

eding-backblaze:
Oh! also, our current practice is to update copyright date on all files that you've updated. So you should probably do that in the files you've touched (not just created) in this PR.

core/src/main/java/com/backblaze/b2/json/B2JsonObjectHandler.java
brianb-backblaze: I'm fine with the change.

brianb-backblaze:
I check the box for "update copyright" in the IDEA commit dialog and have it update automatically.

clamothebb:
Thanks for the reviews. I updated the copyrights. Do the reviewers or the PR author typically perform the merge?

eding-backblaze:
It's typically the PR author that performs the merge. Please choose "Squash and merge" so that all your changes are merged as a single commit into master.

PR #179, opened 2022-11-02
Updates B2WebApiHttpClientImpl to return a subtype of B2Exception on error

nschile:
If the error text is not of type `B2ErrorStructure`, then attempt to return a subtype of B2Exception. Returning a B2Exception subtype enables the `B2Retryer` to retry calls that may succeed on retry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased] - TBD
+### Added
+* Return a subtype of `B2Exception` on errors when the response body does not conform to `B2ErrorStructure`. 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased] - TBD
+### Added
+* Return a subtype of `B2Exception` on errors when the response body does not conform to `B2ErrorStructure`. 
+  Returning a `B2Exception` subtype enables the `B2Retryer` to retry exceptions the may succeed on retry.
```
ab-backblaze, line 6: "the" --> "that"?

ab-backblaze (approved):
Thanks for tracking this down.
It looks like what we talked about, so it still seems good to me.  :)

PR #180, opened 2022-11-09
Teach B2ListFilesIterableBase to handle empty pages when nextFileNam is not null

johnlcox:


eding-backblaze (approved):
Soooo confusing but it looks good to me.

core/src/test/java/com/backblaze/b2/client/B2ListFileNamesIterableTest.java
```diff
@@ -69,6 +71,49 @@ public void testOnePage() throws B2Exception {
...
+    @Test
+    public void testEmptyPageWithNextFileName() throws B2Exception {
+          final B2ListFileNamesRequest request = B2ListFileNamesRequest
+                .builder(BUCKET_ID)
+                .setStartFileName("file/0000")
+                .setMaxFileCount(106)
+                .setPrefix("file/")
+                .setDelimiter("/")
+                .build();
+
```
fmbz, line 83: does it need a test where the first page has results, but the second is empty and has a next file name? (not sure if that would go through a different code path)

fmbz (approved):
LGTM

PR #183, opened 2023-01-10
DP-420 update gradle task checkCode to use 'python3'

lockefox:


fmbz (approved):
LGTM - works in `b2-sdk-java-private` locally

PR #205, opened 2024-12-04
STORAGE-2798 Updated B2StorageClientWebifierImpl methods to improve extensibility

jsha-bbz:
Updated `downloadByName()` and `downloadById()` to use their respective `getDownloadByNameUrl()` and 
`getDownloadByIdUrl()` methods for improved extensibility

### DIFF
--- a/core/src/test/java/com/backblaze/b2/client/B2StorageClientWebifierImplTest.java
+++ b/core/src/test/java/com/backblaze/b2/client/B2StorageClientWebifierImplTest.java
@@ -103,6 +103,7 @@ import static org.junit.Assert.fail;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
@@ -1190,7 +1191,9 @@ public class B2StorageClientWebifierImplTest extends B2BaseTest {
         final B2DownloadByIdRequest request = B2DownloadByIdRequest
                 .builder(fileId(1))
                 .build();
-        webifier.downloadById(ACCOUNT_AUTH, request, noopContentHandler);
+        final B2StorageClientWebifierImpl webifierSpy = spy(webifier);
+        webifierSpy.downloadById(ACCOUNT_AUTH, request, noopContentHandler);
+        verify(webifierSpy).getDownloadByIdUrl(ACCOUNT_AUTH, request);
 
         webApiClient.check("getContent.\n" +
                 "url:\n" +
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/backblaze/b2/client/B2StorageClientWebifierImplTest.java
```diff
@@ -1201,7 +1204,7 @@ public void testDownloadById() throws B2Exception {
                 "    X-Bz-Test-Mode: force_cap_exceeded\n"
         );
 
-        assertEquals(expectedUrl, webifier.getDownloadByIdUrl(ACCOUNT_AUTH, request));
+        assertEquals(expectedUrl, webifierSpy.getDownloadByIdUrl(ACCOUNT_AUTH, request));
```
eding-backblaze, line 1207: why does this need to use the spy'ed object?

core/src/test/java/com/backblaze/b2/client/B2StorageClientWebifierImplTest.java
```diff
@@ -1190,7 +1191,9 @@ public void testDownloadById() throws B2Exception {
         final B2DownloadByIdRequest request = B2DownloadByIdRequest
                 .builder(fileId(1))
                 .build();
-        webifier.downloadById(ACCOUNT_AUTH, request, noopContentHandler);
+        final B2StorageClientWebifierImpl webifierSpy = spy(webifier);
+        webifierSpy.downloadById(ACCOUNT_AUTH, request, noopContentHandler);
+        verify(webifierSpy).getDownloadByIdUrl(ACCOUNT_AUTH, request);
```
eding-backblaze, line 1196: hmm... I think this tests that getDownloadByIdUrl is _called_ but my brain is wondering how useful this verify actually is beyond that. isn't that just an implementation detail? how does verifying it actually tell us more about whether downloadById is working as designed?

jsha-bbz: I was thinking that the combination of the verify and the assertion in the test would test that `downloadById()` is working as designed.

The verify checks that the getDownloadByIdUrl() is being called, which is an implementation detail, but an important one as that was the PR change

This verify alone doesn't test that the method is working properly however along with the assertion that the value from the verify call  (expectedUrl) is the same as the one being passed to the web client call, we test that we call the right thing and that the right thing is passed.

core/src/test/java/com/backblaze/b2/client/B2StorageClientWebifierImplTest.java
jsha-bbz: It doesn't need to be. Is it a good practice to only use the spy reference when using Mockito calls and the non spy in cases like this? I was thinking it might be less confusing if using the same spy reference throughout the code instead of mixing. Don't mind either way. I'll change it back

jsha-bbz:
Closing this PR. As discussed, we don't want clients to create functionality based off of this implementation detail and we will create a better way (pass an http header) to do this

PR #207, opened 2025-05-12
Merge private changes branch

nbehrens-bz:


.github/workflows/deploy.yml
```diff
@@ -12,18 +12,19 @@ jobs:
...
 
-      - name: Use jdk8
-        uses: actions/setup-java@v2
+      - uses: actions/setup-java@v4
         with:
-          distribution: 'adopt'
-          java-version: '8'
-          cache: 'gradle'
+          distribution: 'temurin'
+          java-version: '11'
```
johnlcox, line 22: Does this make the minimum Java version 11 for anyone depending on this library?

nbehrens-bz: No, this is just changing the minimum java version on the worker that runs the ci_cd build.

johnlcox (approved):

